Issue #16, opened 2019-04-02
JSON-RPC: support for getapplicationlog


nimmortal:
Hello.
Could you please add support for getapplicationlog feature.
This is crucial ability to read contract events.

Thanks.

gsmachado:
Hello @nimmortal,
Yes, this is in our pipeline and it will be implemented soon.

nimmortal:
@gsmachado, hello again.

Developing my own contract I reveal, that notification format can be different than simple array. It also may be a map. By the moment such response is not supported by neow3j, but I am planning to add this.

Example:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "txid": "0x8e34291755a55090cb7e7cb5d6c5d24412f91b68f52a90a8f74083f0d72f9c3e",
    "executions": [
      {
        "trigger": "Application",
        "contract": "0x0674dd332fc71c0195efcdcf24300788811c847b",
        "vmstate": "HALT, BREAK",
        "gas_consumed": "2.995",
        "stack": [
          {
            "type": "ByteArray",
            "value": ""
          },
          {
            "type": "ByteArray",
            "value": ""
          },
          {
            "type": "Integer",
            "value": "1"
          }
        ],
        "notifications": [
          {
            "contract": "0xef182f4977544adb207507b0c8c6c3ec1749c7df",
            "state": {
              "type": "Map",
              "value": [
                {
                  "key": {
                    "type": "ByteArray",
                    "value": "67656e6573"
                  },
                  "value": {
                    "type": "ByteArray",
                    "value": "7ee96f2f7592e9de2f5dce5e4df72ddf6abe16ede807fe0d8f636847fe5a1998f98a79bf17a90dce5cac48f91c56364d2d2d5810b300"
                  }
                },
                {
                  "key": {
                    "type": "ByteArray",
                    "value": "62697274685f74696d65"
                  },
                  "value": {
                    "type": "Integer",
                    "value": "1557911235"
                  }
                },
                {
                  "key": {
                    "type": "ByteArray",
                    "value": "636f6f6c646f776e5f656e645f74696d65"
                  },
                  "value": {
                    "type": "ByteArray",
                    "value": ""
                  }
                },
                {
                  "key": {
                    "type": "ByteArray",
                    "value": "6d6f6d5f6964"
                  },
                  "value": {
                    "type": "ByteArray",
                    "value": ""
                  }
                },
                {
                  "key": {
                    "type": "ByteArray",
[...]

PR #20, opened 2019-04-23
#16/applicationlog

nimmortal:
Implement ApplicationLog RPC method described in #16 

Sorry, I made a mistake in commits referencing different issue.


gsmachado:
Thanks a lot for your pull request, @nimmortal! Much appreciated!
We will review the code asap and provide you feedback.

core/src/main/java/io/neow3j/protocol/core/HexUtils.java
```diff
@@ -0,0 +1,31 @@
...
+    public static String hexToString(String input) {
+        return new String(Numeric.hexStringToByteArray(input));
+    }
+
+    public static BigInteger hexToInteger(String input) {
+        String reverse = reverse(input);
+        return Numeric.toBigInt(reverse);
+    }
+
+    public static String scriptHashToAddress(String input) {
```
gsmachado: This is a convenience method that should be placed on `KeyUtils` class. Can you please do it so?
Also, please, move the test to the `KeyUtilsTest` class.

core/src/main/java/io/neow3j/protocol/core/HexUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class HexUtils {
+
+    public static String reverse(String input) {
+        byte[] inputBytes = Numeric.hexStringToByteArray(input);
+        byte[] reversedBytes = ArrayUtils.reverseArray(inputBytes);
+        return Numeric.toHexStringNoPrefix(reversedBytes);
+    }
+
+    public static String hexToString(String input) {
```
gsmachado: This method should be moved to the `Numeric` class, since it's a convenience method for the `Numeric.hexStringToByteArray()`.
Please, also, move the respective test to the `NumericTest` class.

core/src/main/java/io/neow3j/protocol/core/HexUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.neow3j.crypto.KeyUtils;
+import io.neow3j.utils.ArrayUtils;
+import io.neow3j.utils.Numeric;
+
+import java.math.BigInteger;
+
+public class HexUtils {
+
+    public static String reverse(String input) {
```
gsmachado: This method should be moved to the `Numeric` class, even if it's not "numeric" per se. It's just to aggregate all the util methods related to hex, arrays, and others. Also, please, can you move the test to the `NumericTest` class? :-)

gsmachado: Moreover, can you rename this method to `reverseHexString`?

nimmortal:
@gsmachado, thank you for feedback. I tried to fix them all. 

Please, verify my changes.

gsmachado:
After some tests and checking if it works for neo-python and the dotnet cli node, I decided to merge such pull request.

PR #24, opened 2019-05-13
Added raw ClaimTransactions

### LINKED ISSUE #23: Raw ClaimTransactions
As a developer I need to be able to create and send raw transactions of type [ClaimTransactions](https://docs.neo.org/developerguide/en/articles/tx_execution.html#claimtransaction).

This does not yet include automatic retrieval or tracking of claimable GAS.

claudemiller:
Extended RawTransaction and implemented ClaimTransaction. This resolves #23.

### DIFF
--- a/model/src/main/java/io/neow3j/model/types/TransactionType.java
+++ b/model/src/main/java/io/neow3j/model/types/TransactionType.java
@@ -4,27 +4,29 @@ import com.fasterxml.jackson.annotation.JsonValue;
 
 public enum TransactionType {
 
-    MINER_TRANSACTION("MinerTransaction", 0x00),
-    ISSUE_TRANSACTION("IssueTransaction", 0x01),
-    CLAIM_TRANSACTION("ClaimTransaction", 0x02),
-    DATA_FILE("DataFile", 0x12),
-    ENROLLMENT_TRANSACTION("EnrollmentTransaction", 0x20),
-    REGISTER_TRANSACTION("RegisterTransaction", 0x40),
-    CONTRACT_TRANSACTION("ContractTransaction", 0x80),
-    RECORD_TRANSACTION("RecordTransaction", 0x81),
-    STATE_TRANSACTION("StateTransaction", 0x90),
-    STATE_UPDATE_TRANSACTION("StateUpdateTransaction", 0x90),
-    STATE_UPDATER_TRANSACTION("StateUpdaterTransaction", 0x91),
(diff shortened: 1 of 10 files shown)


model/src/main/java/io/neow3j/model/types/TransactionType.java
```diff
@@ -4,27 +4,29 @@
...
+    DESTROY_TRANSACTION("DestroyTransaction", 0x18, 0),
+    PUBLISH_TRANSACTION("PublishTransaction", 0xd0, 0),
+    INVOCATION_TRANSACTION("InvocationTransaction", 0xd1, 0);
 
     private String jsonValue;
     private byte byteValue;
+    private byte version;
 
-    TransactionType(String jsonValue, int v) {
+    TransactionType(String jsonValue, int v, int version) {
```
gsmachado: The version was added here because, maybe, other transaction types could come up with other versions? If yes, then it's fine.

utils/src/main/java/io/neow3j/utils/Numeric.java
```diff
@@ -79,8 +79,8 @@ public static boolean containsHexPrefix(String input) {
                 && input.charAt(0) == '0' && input.charAt(1) == 'x';
     }
 
-    public static BigDecimal toBigDecimal(byte[] value) {
-        return new BigDecimal(toBigInt(value));
+    public static BigDecimal fromFixed8ToBigDecimal(byte[] value) {
```
gsmachado: Cool that you found that. :-) Maybe it was even affecting other functionalities.

gsmachado:
There are no significant things to be further enhanced in this pull request. Good one. 👍

model/src/main/java/io/neow3j/model/types/TransactionType.java
claudemiller: Yes that is the case. Transaction types could be different in version. They don't necessarily change versions all at the same time

utils/src/main/java/io/neow3j/utils/Numeric.java
claudemiller: The method is not used anywhere else. But i did not yet check for similar methods that might have the same problem.

gsmachado: Alright. Please, add to your ToDo list to check that out soon (i.e., similar methods with the same problem).

PR #31, opened 2019-05-17
changed RawTransactionInput prevIndex to big-endian

shoxive:
A serialized transaction must serialize utxo's prevIndex in big-endian,or The transaction will failed unless it's input index was zero.Please confirm this commit.


gsmachado:
Thanks for the pull request. Yes, this makes sense.

gsmachado:
Usually, I request a test case for each pull request before accepting. But this time I already have a test case for it and I will just extend it to cover such case.

Issue #38, opened 2019-05-31
JSON-RPC: getclaimable method


claudemiller:
As a developer, I would like to use the getclaimable JSON-PRC call to easily fetch transaction outputs for a specific NEO address that allow me to claim GAS.

https://docs.neo.org/en-us/node/cli/latest-version/api/getclaimable.html

claudemiller:
Only integration tests missing.

claudemiller:
Integration test covered by issue #94

PR #39, opened 2019-05-31
Implementation of getclaimable method and tests.

claudemiller:
Resolves issue #38.


gsmachado:
It seems good to me. :-)

gsmachado:
I also checked codecov, and the test coverage also did not get worse -- which is a good sign. :-)

PR #43, opened 2019-06-11
#16: implement applicationlog map and array support

nimmortal:
@gsmachado I made an example how it may be, to be honest I am not sure that it is the best solution, but there are a lot questions, especially about the formats. As a result I not fully covered code with tests.


gsmachado:
Thanks for the pull request, @nimmortal! I'll review it asap.

nimmortal:
Looks like Map support should be also in regular invocations. Unfortunately, I did not found any descriptions in docs.

gsmachado:
@nimmortal why did you close?
Sorry for not getting back in a timely manner.

nimmortal:
@gsmachado Sorry, I have had to rename branch to use it thought jitpack repository in my companies project. I will sync branch with current master and create one more PR.

PR #44, opened 2019-06-13
Implementation of listplugins RPC method.

### LINKED ISSUE #42: JSON-RPC: listplugins method
As a developer, I would like to use the listplugins JSON-PRC call to check what plugins a node supports. This call became available with neo-cli 2.10.1.

https://docs.neo.org/en-us/node/cli/2.10.1/api/listplugins.html

claudemiller:
Resolves #42


gsmachado:
Looks good to me. :-)

Issue #29, opened 2019-05-15
wallet: transfer NEO and GAS assets


gsmachado:
As a developer, I would like to have methods exposed in the `wallet` module (#25) that I could:
* transfer NEO and GAS assets to any NEO address

NEO addresses should be validated before performing the transaction.

gsmachado:
This feature is available since neow3j version [2.0.0](https://github.com/neow3j/neow3j/releases/tag/2.0.0). Check the [releases page](https://github.com/neow3j/neow3j/releases) for the newest version.

Issue #27, opened 2019-05-15
wallet: get balances (NEO and GAS)


gsmachado:
As a developer, I would like to have methods exposed in the `wallet` module (#25) where I could get balances for NEO and GAS assets -- for addresses already existing on the wallet.

The UTXOs should be calculated based on the addresses added to the wallet, and the local file-based database state.

claudemiller:
Issue was resolved with pull-request #45.
The Account class has a `balance` variable which holds all balances of that account. The implementation hands the control over when to update the balances to the developer. This is because an update requires an RPC. With `updateAssetBalances()` and `updateTokenBalances()` the developer is able to update the balances when he sees fit.

PR #45, opened 2019-06-20
Wallet, Account, Balances and Asset Transfer implementation

### LINKED ISSUE #26: wallet: create, import, recover, and export
As a developer, I would like to have methods on the `wallet` module (convenience layer, discussed by #25) that I could:
* create new addresses to a wallet (not only NEP-6 and NEP-2, but also with BIP-39)
* import a wallet/address (NEP-6, NEP-2, .json format)
* recover a wallet/address (NEP-6, NEP-2, BIP-39, .json format)
* export (backup) a wallet/address (NEP-6, NEP-2, .json format)

Most of these functionalities are kind of wrappers (abstraction layer) to the methods implemented on the `core` and/or `crypto` modules.

claudemiller:
Resolves #26, #27, #29.


wallet/src/main/java/io/neow3j/wallet/AssetTransfer.java
```diff
@@ -0,0 +1,227 @@
...
+            }
+            byte[] rawUnsignedTx = tx.toArray();
+            ECKeyPair keyPair = account.getECKeyPair();
+
+            RawInvocationScript invocationScript = new RawInvocationScript(
+                    Sign.signMessage(rawUnsignedTx, keyPair));
+
+            String publicKeyHexString = Numeric.toHexStringNoPrefix(account.getPublicKey());
+            byte[] publicKeyByteArray = Numeric.hexStringToByteArray(publicKeyHexString);
+            RawVerificationScript verificationScript = Keys.getVerificationScriptFromPublicKey(publicKeyByteArray);
```
gsmachado: Simplify?! :-)

wallet/src/main/java/io/neow3j/wallet/Balances.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Overrides the asset balances with the provided information in the unspents.
+     * @param unspents The unspents fetched from a RPC node.
+     */
+    public void updateAssetBalances(NeoGetUnspents.Unspents unspents) {
+        if (!unspents.getAddress().equals(account.getAddress())) {
+            throw new IllegalArgumentException("The provided unspents are not related to the " +
+                    "account of this balance");
+        }
+        assetBalances = new HashMap<>();
```
gsmachado: Can we infer the assetBalances size? ;-)

wallet/src/main/java/io/neow3j/wallet/Balances.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Overrides the token balances with the provided balances.
+     * @param balances The NEP5 token balances fetched from a RPC node.
+     */
+    public void updateTokenBalances(NeoGetNep5Balances.Balances balances) {
+        if (!balances.getAddress().equals(account.getAddress())) {
+            throw new IllegalArgumentException("The provided token balances are not related to the " +
+                    "account of this balance");
+        }
+        this.tokenBalances = new HashMap<>();
```
gsmachado: Infer the size?! Same as above.

wallet/src/test/java/io/neow3j/wallet/WalletTest.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Test
+    public void testCreateWalletFromNEP6File() throws IOException {
+
+        URL nep6WalletFile = Thread.currentThread().getContextClassLoader().getResource("wallet.json");
+        ObjectMapper mapper = new ObjectMapper();
+        NEP6Wallet nep6Wallet = mapper.readValue(nep6WalletFile, NEP6Wallet.class);
+
+        Wallet w = Wallet.fromNEP6Wallet(nep6Wallet).build();
```
gsmachado: As discussed, create a method called "load" or something like this, that will get as parameter a String or Path with the path to the wallet (.json).

wallet/src/main/java/io/neow3j/wallet/Account.java
```diff
@@ -0,0 +1,274 @@
...
+            this.encryptedPrivateKey= NEP2.encrypt(password, getECKeyPair(), scryptParams);
+        }
+    }
+
+    public boolean isMultiSig() {
+        if (contract != null && contract.getScript() != null) {
+            String script = contract.getScript();
+            return script.substring(script.length() - 2).equals(OpCode.toHexString(CHECKMULTISIG));
+        }
+        return false;
```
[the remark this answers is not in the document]
claudemiller: I can currently not determine if it is a multi-sig if the verification script is not given. We could allow an account to hold more than one public key, easily identifying it as a multi-sig account. 
But I wish to first get feedback from developers or the NGF about usage of multi-sig accounts. Maybe we don't need to spend so much though on it.

wallet/src/test/java/io/neow3j/wallet/WalletTest.java
claudemiller: I added a method that takes an URL object. I guess usually one gets a resource via `SomeClass.class.getResource("some_resource")` which returns an URL.

gsmachado:
It's fine now. :-) Great work!

gsmachado:
It's fine now. :-) Great work!

PR #49, opened 2019-06-25
Convenience methods for AssetTransfer

claudemiller:
Added methods for constructing an asset transfer with a single output given by assetId, amount, and receiving account. Added related tests. Related to #29.

### DIFF
--- a/wallet/src/test/java/io/neow3j/wallet/AssetTransferTest.java
+++ b/wallet/src/test/java/io/neow3j/wallet/AssetTransferTest.java
@@ -18,7 +18,6 @@ import java.math.BigDecimal;
 import java.util.Arrays;
 
 import static junit.framework.TestCase.assertEquals;
-import static junit.framework.TestCase.fail;
 
 public class AssetTransferTest {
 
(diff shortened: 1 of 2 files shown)


wallet/src/test/java/io/neow3j/wallet/AssetTransferTest.java
```diff
@@ -122,6 +122,196 @@ public void test_transfer_with_fee() throws IOException, ErrorResponseException
...
+                    .amount(BigDecimal.ONE)
+                    .toAddress("AJQ6FoaSXDFzA6wLnyZ1nFN7SGSN2oNTc3")
+                    .build();
+        } catch(IllegalStateException e) {
+            return;
+        }
+        fail();
+    }
+
+    @Test
```
gsmachado: Can you please adapt all the tests that you expect an exception with:
@Test(expect = SpecificException.class)

wallet/src/test/java/io/neow3j/wallet/AssetTransferTest.java
```diff
@@ -122,6 +122,196 @@ public void test_transfer_with_fee() throws IOException, ErrorResponseException
...
+                    .asset(NEOAsset.HASH_ID)
+                    .output(output)
+                    .build();
+        } catch(IllegalStateException e) {
+            return;
+        }
+        fail();
+    }
+
+    @Test
```
gsmachado: same here!
The catch clause is not required.

wallet/src/test/java/io/neow3j/wallet/AssetTransferTest.java
claudemiller: Done

wallet/src/test/java/io/neow3j/wallet/AssetTransferTest.java
claudemiller: Done

PR #59, opened 2019-07-04
#52 RawScript, InvocationScript, VerificationScript rework

### LINKED ISSUE #52: The invocation and verification in RawScript should be more flexible to support contract
The invocation and verification in RawScript should be more flexible to support contract
Here is a simply example for this requirement

image
image

claudemiller:
Verification and Invocation scripts now hold arbitrary byte arrays. 
This change required many other changes and new tests which are all part of this pull-request.

Resolves #52


crypto/src/test/java/io/neow3j/crypto/transaction/RawInvocationScriptTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    public void testFromMessageAndKeyPair() throws InvalidAlgorithmParameterException,
+            NoSuchAlgorithmException, NoSuchProviderException {
+
+        byte[] message = new byte[10];
+        Arrays.fill(message, (byte) 10);
+        ECKeyPair keyPair = Keys.createEcKeyPair();
+        RawInvocationScript invScript = RawInvocationScript.fromMessageAndKeyPair(message, keyPair);
+        SignatureData expectedSignature = Sign.signMessage(message, keyPair);
+        byte[] expectedScript = ArrayUtils.concatenate(PUSHBYTES64.getValue(), expectedSignature.getConcatenated());
```
gsmachado: Well, I prefer to use a String with hexadecimal values and commenting them on the test itself -- for a better understanding.
E.g.:
```
byte[] expectedScript = Numeric.hexStringToByteArray(
    // PUSHBYTES64
    "64" +
    // signature
    "00000000000000000000000"
);
```

But it's a matter of style as well. The disadvantage is that such string can get pretty large. The advantage is a better understanding of how things are built.
Anyway, testing and making it work is more important. 👍 :-)

crypto/src/test/java/io/neow3j/crypto/transaction/RawScriptTest.java
```diff
@@ -0,0 +1,182 @@
...
+import static io.neow3j.constants.OpCode.CHECKMULTISIG;
+import static io.neow3j.constants.OpCode.CHECKSIG;
+import static io.neow3j.constants.OpCode.PUSH2;
+import static io.neow3j.constants.OpCode.PUSH3;
+import static io.neow3j.constants.OpCode.PUSHBYTES33;
+import static io.neow3j.constants.OpCode.PUSHBYTES64;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+public class RawScriptTest {
```
gsmachado: I loved to read this class! ❤️ 💯 🥇

utils/src/main/java/io/neow3j/constants/NeoConstants.java
```diff
@@ -15,4 +14,11 @@
 
     public static final byte COIN_VERSION = 0x17;
 
+    /**
+     * The maximum number of public keys that can take part in a multi-signature address.
+     * Taken from Neo.SmartContract.Contract.CreateMultiSigRedeemScript(...) in the C# neo repo
+     * at https://github.com/neo-project/neo.
+     */
+    public static final int MAX_PUBLIC_KEYS_PER_MULTISIG_ACCOUNT = 1024;
```
gsmachado: I didn't know about that! Well-done. 💯 👍 🎉

core/src/test/java/io/neow3j/protocol/transaction/ContractTransactionTest.java
```diff
@@ -33,15 +38,13 @@
...
-                .outputs(Arrays.asList(
-                        new RawTransactionOutput(NEOAsset.HASH_ID, "10.0", "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"),
-                        new RawTransactionOutput(NEOAsset.HASH_ID, "90.0", "AKYdmtzCD6DtGx16KHzSTKY8ji29sMTbEZ")
-                ))
-                .scripts(Arrays.asList(new RawScript(
-                        Arrays.asList(),
-                        new RawVerificationScript(Arrays.asList(Numeric.toBigIntNoPrefix("0265bf906bf385fbf3f777832e55a87991bcfbe19b097fb7c5ca2e4025a4d5e5d6")), 1))))
+                .output(new RawTransactionOutput(NEOAsset.HASH_ID, "10.0", "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"))
+                .output(new RawTransactionOutput(NEOAsset.HASH_ID, "90.0", "AKYdmtzCD6DtGx16KHzSTKY8ji29sMTbEZ"))
+//                .script(new RawScript(new byte[]{}, RawVerificationScript.fromPublicKey(publicKey).getScript()))
```
gsmachado: Why is this commented? Review it.

crypto/src/main/java/io/neow3j/crypto/transaction/RawInvocationScript.java
```diff
@@ -1,72 +1,115 @@
...
+        SignatureData signature = Sign.signMessage(message, keyPair);
+        return fromSignature(signature);
     }
 
-    public Sign.SignatureData getSignature() {
-        return signature;
+    public static RawInvocationScript fromSignatures(List<SignatureData> sigs) {
+        ByteBuffer buffer = ByteBuffer.allocate(sigs.size() + sigs.size()*PUSHBYTES64.getValue());
+        for (SignatureData sig : sigs) {
+            buffer.put(PUSHBYTES64.getValue());
```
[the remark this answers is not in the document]
claudemiller: As far as I know the signatures in NEO are always 64 bytes long. But, you are right to question this. When I replace this code part with ScriptBuilder, the PUSHBYTES will not be set statically anymore, but rather dynamically according to the real length of the signature.

claudemiller:
> In general, it looks good. 👍 💪
> 
> My main concern is in the testing part: we would need to have a more "real-test" set-up, where we compare serialized transactions (hex) accepted on a privatenet to what `neow3j` generates. Thus, we can make sure that what we generate is correct in practical terms.

We have several tests in ContractTransactionTest and ClaimTransactionTest which use byte arrays taken from real, "working" transactions. With those the correctness of the RawScript is also covered :)

PR #60, opened 2019-07-08
Extract script building code into a ScriptBuilder class

claudemiller:
Script building and serialization should not be mixed.
Therefore the script building functionality is separated into its own
ScriptBuilder class. RawVerificationScript and RawInvocationScript
now use the ScriptBuilder for constructing themselves.
Also added more OpCodes of which only some are currently in use.


gsmachado:
So far, I think the ScriptBuilder is good. However, before merging to `develop` I would implement some unit tests to confirm the correctness of such script builder (even if the tests involving the RawVerificationScript and RawInvocationScript are still passing).

Issue #57, opened 2019-07-04
Wallet: method for creating a new wallet and account


gsmachado:
As a developer, I would like to have a method (maybe, static) on the `Wallet` class to create a wallet with an account already on it.

For example, like this a wallet object is created with an account already added to it:

```
Wallet w1 = Wallet.create();
```

Or, you can create a wallet object and already persist to a NEP6-compatible .json file, like:

```
Wallet w1 = Wallet.create("/tmp/wallet.json");
```

claudemiller:
The above commits resolve the first part of the issue but not the second (creation with writing the wallet to a file immediately).

PR #61, opened 2019-07-15
#57: wallet enhancements

gsmachado:


wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -164,15 +173,73 @@ public static Builder fromNEP6Wallet(NEP6Wallet nep6Wallet) {
         return b;
     }
 
+    /**
+     * Creates a NEP6 compatible wallet file.
+     *
+     * @param destination the file or directory that the wallet file should be saved.
+     *                    If a directory is provided, then a filename will be automatically chosen.
```
claudemiller: Here it says that if the destination param is a directory a file will still be created with automatic file name.
But in the method body an exception is thrown if the param is not a file.

gsmachado: Good catch. Such comment remained untouched after a refactor. Fixed.

PR #64, opened 2019-07-19
Different serialization needed for sending or signing a transaction

claudemiller:


crypto/src/main/java/io/neow3j/crypto/transaction/RawTransaction.java
```diff
@@ -121,15 +116,41 @@ private void serializeWithoutScripts(BinaryWriter writer) throws IOException {
...
-        }
+        writer.writeSerializableVariable(this.scripts);
     }
 
     public abstract void serializeExclusive(BinaryWriter writer) throws IOException;
 
     public abstract void deserializeExclusive(BinaryReader reader) throws IOException, IllegalAccessException, InstantiationException;
 
+    /**
+     * Serializes this transaction to a raw byte array without any scripts. This is the required
```
gsmachado: Minor change. It should be written: "This is required if it's the serialized transaction gets signed, e.g., by an external keypair/provider."

PR #73, opened 2019-07-24
#16: implement applicationlog map and array support

nimmortal:


nimmortal:
@gsmachado as I told you before I created a new branch due to jitpack. Please, take a look at my changes. My main point is that new entity 'Item' / 'StackItem' is a bit different and inherits ContractParameter, this is needed because as contract output maybe complex structures as Array and Map. Especially Map, because this is not supported as input.

nimmortal:
@gsmachado, @claudemiller Guys, any progress on that?

claudemiller:
Hey @nimmortal,
I'll go through your changes asap. 
Good point with the new `StackItem` entity.

claudemiller:
Hey @nimmortal 
I created this [pull-request](https://github.com/nimmortal/neow3j/pull/2) to your forked repo with a bit of an alternative approach to the StackItem.
Please check it out and give me feedback.

nimmortal:
@claudemiller I solved conflicts with current master, I saw there was changes according to the old implementation, I decided to take the new model how you done. 

@claudemiller @gsmachado can you please review this changes and merge it to the master?

core/src/main/java/io/neow3j/protocol/core/StackItemParser.java
```diff
@@ -0,0 +1,46 @@
+package io.neow3j.protocol.core;
+
+import io.neow3j.model.types.StackItemType;
+import io.neow3j.protocol.core.methods.response.StackItem;
+import io.neow3j.utils.Keys;
+import io.neow3j.utils.Numeric;
+
+import java.math.BigInteger;
+
```
claudemiller: The StackItemParser is not really needed anymore as a separate class.
The methods are almost exclusively used for the `BYTE_ARRAY` type. I.e. they can be moved directly to the `ByteArrayStackItem`class.

core/src/main/java/io/neow3j/protocol/deserializer/StackDeserializer.java
```diff
@@ -0,0 +1,92 @@
...
+
+        JsonNode valueNode = param.get("value");
+        if (valueNode != null) {
+            switch (type) {
+                case BYTE_ARRAY:
+                    return new ByteArrayStackItem(valueNode.asText());
+                case BOOLEAN:
+                    return new BooleanStackItem(valueNode.asBoolean());
+                case INTEGER:
+                    return new IntegerStackItem(new BigInteger(valueNode.asText()));
```
claudemiller: You mentioned that in case of a zero-valued integer the text value here will be '', i.e. the empty string. 
Please add a check here which passes a BigInteger.ZERO to the constructor in case of the empty string.

core/src/main/java/io/neow3j/protocol/core/methods/response/ArrayStackItem.java
```diff
@@ -0,0 +1,23 @@
+package io.neow3j.protocol.core.methods.response;
+
+import io.neow3j.model.types.StackItemType;
+
+import java.util.List;
+
+public class ArrayStackItem extends StackItem {
```
claudemiller: As you mentioned we must add a `from(StackItem stackItem)` method to every `StackItem` subtype. This will improve the readability for the user substantially when he wants to cast a `StackItem` to a specific subtype.

nimmortal:
@claudemiller Unfortunately, now we have strict deadline. I assume that I can return to this after a week. If it is ok for you, then I can make changes in PR.

PR #77, opened 2019-07-30
Replace plane byte arrays with ScriptHash class

### LINKED ISSUE #75: The little-endian and big-endian of scriptHash is not consistent
The ContractInvocation uses big-endian contractScriptHash as parameter, at the same time the address of a smart contrtact should be encoded with little-endian scriptHash

```
    public static String scriptHashToAddress(String input) {
        byte[] inputBytes = Numeric.hexStringToByteArray(input);
        return toAddress(inputBytes);
    }
```

claudemiller:
We want more clarity when it comes to handling script hashes. That's why we now use a ScriptHash class that clearly handles the endianness of script hashes.
At the same time the ContractParameter class was changed and now uses the value types according to their parameter type instead of just strings. E.g. a byte array parameter now holds value of type byte[].

Resolves #75


gsmachado:
@claudemiller before anything, can you check Codacy and try to go over the points where it makes sense?

PR #88, opened 2019-08-07
Feature/contract deploy

gsmachado:
Please, check the whole code and maybe think if we need to have more tests -- mainly integration tests.


gsmachado:
Wow! +2.23% of test coverage!
And it could go higher if we had more tests for the `NeoABIUtils` and `ContractDeployment` classes.
Anyway, nice!

contract/src/main/java/io/neow3j/contract/ContractFunctionProperties.java
```diff
@@ -153,12 +120,14 @@ public void deserialize(BinaryReader reader) throws IOException {
             this.needsDynamicInvoke = unpackNeedsDynamicInvoke(functionProperties);
             this.isPayable = unpackIsPayable(functionProperties);
             int returnTypeInt = reader.readPushInteger();
-            this.returnType = toSerializable(ContractParameterType.valueOf((byte) returnTypeInt));
-
+            this.returnType = ContractParameterType.valueOf((byte) returnTypeInt);
             // TODO: 2019-06-04 Guil:
```
claudemiller: This comment has to be addressed. I think it is an actual issue when deserializing parameter types like `InteropInterface`.

contract/src/main/java/io/neow3j/contract/ContractAbiLoader.java
```diff
@@ -58,8 +75,8 @@ public Builder contractScriptHash(ScriptHash contractScriptHash) {
             return this;
         }
 
-        public Builder address(String address) {
-            this.contractScriptHash = ScriptHash.fromAddress(address);
+        public Builder address(String contractScriptHash) {
+            this.contractScriptHash = new ScriptHash(Numeric.cleanHexPrefix(contractScriptHash));
             return this;
         }
```
claudemiller: In other places we "force" the user to explicitly use the ScriptHash class instead of raw strings or byte arrays. The ScriptHash class centralizes the endianness issue. If we use the ScriptHash class we don't need to mention the required endianness in every method that deals with script hashes.

PR #89, opened 2019-08-08
Update and rework core integration tests

claudemiller:
A rework of the already existing integration tests. Some had to be adapted because the private net container changed. They are now split into read-only and write tests to save some time. The read-only tests all run on the same container instance.


claudemiller:
The remaining issues in Codacy will not be fixed. The empty tests are there as a reminder for which RPC methods we should still write integration tests for.

PR #91, opened 2019-08-20
Replace ContractParameters with StackItems in stack information returned by RPC nodes

claudemiller:


.travis.yml
```diff
@@ -1,3 +1,4 @@
+dist: trusty
```
gsmachado: what was the problem here?

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
```diff
@@ -0,0 +1,187 @@
...
+            if (typeNode != null) {
+                type = jp.getCodec().treeToValue(typeNode, StackItemType.class);
+            }
+            if (valueNode == null) {
+                return new StackItem(type, null);
+            }
+            if (type == null) {
+                return new StackItem(null, valueNode.asText());
+            }
+            switch (type) {
```
gsmachado: Yeah, this is a bit nasty, but we can live with it... I try to not use deal with JSON serialization/deserialization as much as possible, but sometimes we can't avoid to use it...

core/src/main/java/io/neow3j/protocol/core/methods/response/ArrayStackItem.java
```diff
@@ -0,0 +1,38 @@
...
+     * Gets the stack item at the given index of this stack item array.
+     *
+     * @param i index of the stack item to return
+     * @return the stack item at the given index
+     */
+    public StackItem get(int i) {
+        return getValue().get(i);
+    }
+
+    public int size() {
```
gsmachado: I know it's a bit painful (and sometimes does not even make sense), but could you add JavaDocs for all public methods? Even if it's just a few words.

core/src/main/java/io/neow3j/protocol/core/methods/response/ArrayStackItem.java
```diff
@@ -0,0 +1,38 @@
...
+        return getValue().size();
+    }
+
+    public boolean isEmpty() {
```
gsmachado: here as well.

core/src/main/java/io/neow3j/protocol/core/methods/response/MapStackItem.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param key the key whose associated value is to be returned.
+     * @return the value to which the given key is mapped, or null if this map stack item
+     * contains no mapping for the key.
+     */
+    public StackItem get(StackItem key) {
+        return getValue().get(key);
+    }
+
+    public int size() {
```
gsmachado: Annotation @Override is missing?

core/src/main/java/io/neow3j/protocol/core/methods/response/MapStackItem.java
```diff
@@ -0,0 +1,61 @@
...
+        return getValue().size();
+    }
+
+    public boolean isEmpty() {
```
gsmachado: here as well?

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
```diff
@@ -0,0 +1,187 @@
...
+
+    protected StackItemType type;
+    protected Object value;
+
+    public StackItem(StackItemType type, Object value) {
+        this.type = type;
+        this.value = value;
+    }
+
+    public StackItemType getType() {
```
gsmachado: JavaDocs?! Makes sense to add it here?

I'm not sure if JavaDocs from an extended class (with @Override annotation) will inherit the JavaDocs as well, or not? Check that.

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Override
+    public String toString() {
+        return "StackItem{" +
+                "type=" + this.type +
+                ", value=" + this.value +
+                '}';
+    }
+
+    public ByteArrayStackItem asByteArray() {
```
gsmachado: JavaDocs for all these convenience methods.

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public StructStackItem asStruct() {
+        if (this instanceof StructStackItem) {
+            return (StructStackItem) this;
+        }
+        throw new IllegalStateException("This stack item is not of type " +
+                StackItemType.STRUCT.jsonValue() + " but of " + this.type.jsonValue());
+    }
+
+    public static class StackDeserializer extends StdDeserializer<StackItem> {
```
gsmachado: I think that documentation on the serialization/deserialization does not make sense. So, leave it. :-)

gsmachado:
The comments I've made are not critical. It's just about documentation and convenience methods. If you judge that my comments are not important at all and should not be implemented, let me know.

.travis.yml
claudemiller: See for example this build https://travis-ci.org/neow3j/neow3j/builds/574232995.
I got this fix from the following thread: https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766

core/src/main/java/io/neow3j/protocol/core/methods/response/MapStackItem.java
claudemiller: I didn't make the `MapStackItem` implement the `Map` interface. Otherwise, I would have to implement many more methods. I just added a few convenience methods. For all other map-related methods the user will have to call `getValue()` and then call the methods on the actual `Map` directly.

core/src/main/java/io/neow3j/protocol/core/methods/response/MapStackItem.java
claudemiller: Same as above

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
claudemiller: Yes, javadoc is inherited on methods that bear the @Override annotation.

gsmachado:
Thanks for the changes. I think it's ready to be merged.

PR #99, opened 2019-11-05
NEP5 contract support

Priyanshu-soni:
includes : 
  - totalSupply
  - name
  - symbol
  - decimals

### DIFF
--- a/wallet/src/main/java/io/neow3j/wallet/Nep5.java
+++ b/wallet/src/main/java/io/neow3j/wallet/Nep5.java
@@ -8,7 +8,7 @@ import java.io.IOException;
 import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
+import java.util.Optional;
 
 public class Nep5 {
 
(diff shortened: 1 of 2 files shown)


gsmachado:
Hey @Priyanshu-soni, thanks for the PR! It's good progress, but I have some things we need to clarify before -- which I'll address in the review.

wallet/src/main/java/io/neow3j/wallet/Nep5.java
```diff
@@ -0,0 +1,85 @@
...
+public class Nep5 {
+    private Neow3j neow3j;
+    private ScriptHash fromContractScriptHash;
+
+    private Nep5(Builder builder) {
+        this.neow3j = builder.neow3j;
+        this.fromContractScriptHash = builder.fromContractScriptHash;
+    }
+
+
```
gsmachado: You should use a single line space between methods. Please, change it for all the files you've committed.

wallet/src/main/java/io/neow3j/wallet/Nep5.java
```diff
@@ -0,0 +1,85 @@
...
+        List<ContractParameter> params = Arrays.asList(
+                ContractParameter.string("totalSupply"),
+                ContractParameter.array());
+        NeoInvoke send = neow3j.invoke(this.fromContractScriptHash.toString(), params).send();
+        return send.getInvocationResult().getStack().get(1).asByteArray().getAsNumber();
+    }
+
+
+    public String name() throws IOException {
+
```
gsmachado: Unused spaces between method declaration and the start of the method's code. Please change in this file and all files of this PR.

wallet/src/test/java/io/neow3j/wallet/Nep5Test.java
```diff
@@ -0,0 +1,52 @@
...
+import java.math.BigInteger;
+
+public class Nep5Test {
+
+    private Neow3j neow3j;
+    private ScriptHash contractScriptHash;
+
+    @Before
+    public void setUp() {
+        this.neow3j = Neow3j.build(new HttpService("http://localhost:30333"));
```
gsmachado: Currently, the tests are failing on Travis. This is because you're using "localhost:30333" as the Neo node.
You **must** mock the neow3j behavior to only test the Nep5 functionality, and not other code abstraction layers.
Please check the class `ContractDeploymentTest.class` for an example on how to mock and test higher layers that rely on the `Neow3j` class.

wallet/src/test/java/io/neow3j/wallet/Nep5Test.java
```diff
@@ -0,0 +1,52 @@
...
+import io.neow3j.protocol.Neow3j;
+import io.neow3j.protocol.http.HttpService;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.math.BigInteger;
+
+public class Nep5Test {
```
gsmachado: You should not only test the "happy path", but also exceptions or things that can go wrong. Please include tests for that.

Priyanshu-soni:
Hey @gsmachado , please check the updated PR

wallet/src/main/java/io/neow3j/wallet/Nep5.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public static class Builder {
+        private Neow3j neow3j;
+        private ScriptHash fromContractScriptHash;
+        public Builder(Neow3j neow3j) {
+            this.neow3j = neow3j;
+        }
+
+        public Nep5.Builder fromContract(ScriptHash contractScriptHash) {
```
gsmachado: @Priyanshu-soni could we, somehow, check whether this contract is a Nep5 standard using the JSON-RPC `invoke`?
@csmuller what's your opinion? Should we do this check here? Or an exception would be thrown anyway once a method that does not exist is called in this contract?

wallet/src/test/java/io/neow3j/wallet/Nep5Test.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.io.IOException;
+import java.math.BigInteger;
+
+import static junit.framework.TestCase.assertEquals;
+import static junit.framework.TestCase.fail;
+
+public class Nep5Test {
```
gsmachado: @Priyanshu-soni in the report of your last commit (https://github.com/neow3j/neow3j/pull/99#issuecomment-556003123), the test coverage decreased by 0.3%. We always should keep the same coverage level or increase with pull requests. Can you check which part of the code that is not covered, somehow?

wallet/src/test/java/io/neow3j/wallet/Nep5Test.java
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals("Nep5 test for Symbol", nep5.symbol());
+    }
+
+    @Test
+    public void decimalsTest() throws IOException {
+        Mockito.doReturn(new BigInteger("66554545")).when(nep5).decimals();
+        assertEquals(new BigInteger("66554545"), nep5.decimals());
+    }
+
+    @Test
```
gsmachado: @Priyanshu-soni when you're expecting an exception, can you please use the "expected" attribute in the `@Test` annotation? (instead of doing a try-catch and then fail)
Like here: https://junit.org/junit4/javadoc/latest/org/junit/Test.html

wallet/src/main/java/io/neow3j/wallet/Nep5.java
```diff
@@ -0,0 +1,75 @@
...
+        NeoInvoke send = neow3j.invoke(this.fromContractScriptHash.toString(), params).send();
+        return send.getInvocationResult().getStack().get(1).asByteArray().getAsString();
+    }
+
+    public BigInteger decimals() throws IOException {
+        List<ContractParameter> params = Arrays.asList(
+                ContractParameter.string("decimals"),
+                ContractParameter.array());
+        NeoInvoke send = neow3j.invoke(this.fromContractScriptHash.toString(), params).send();
+        return (BigInteger) send.getInvocationResult().getStack().get(1).getValue();
```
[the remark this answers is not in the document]
csmuller: At the moment we have to use `send.getInvocationResult().getStack().get(1)`.
The only thing we can change here is use `asInteger().getValue()` instead of just `getValue()` and then casting it to BigInteger.

wallet/src/main/java/io/neow3j/wallet/Nep5.java
csmuller: I don't think we should check the contract for every NEP-5 method. With the possibilities in Neo 2 this would imply doing `invoke`s for every method.
Though, there is a new standard, NEP-10, which defines that a smart contract should have a method `supportedStandards` that would tell us that the contract supports the NEP-5 API.
For now we can ignore that and leave it to the developer to decide if the wrapped contract fulfills the NEP-5 standard. I will keep track of NEP-10.

gsmachado:
Hello @Priyanshu-soni: it seems that the codacy code quality is not passing, somehow. Can you check that before I go through another round of review?

PR #102, opened 2019-12-07
Neo 3 transaction model

### LINKED ISSUE #97: Adapt transaction model to Neo 3
The transaction model has changed in Neo 3. The transaction types have been unified, so that only one transaction type exists. The data structure has also changed considerably. See the [Neo3 Dev Guide](https://github.com/neo-ngd/NEO3-Development-Guide/tree/master/en/Transactions#transaction-structure) for more information on the changes.

- [x] Adapt `Cosigner`, `TransactionAttribute`, `TransactionAttributeUsageType`, and `Transaction` to newest changes in `neo` core.

csmuller:
Created classes for the new transaction model. The old model is still in place and the high level classes like `AssetTransfer` and `ContractInvocation` still depend on the old model. I didn't adapt them because they have to be replaced under the new transaction model.

Many files are only changed because of name changes (e.g. `RawTransactionAttribute` to `TransactionAttribute`). And another set of files are only changed because of the addition of the `DeserializationException` that replaces all the exceptions that can appear in the deserialization.

Resolves #97


core/src/main/java/io/neow3j/transaction/Transaction.java
```diff
@@ -0,0 +1,463 @@
...
+     */
+    public byte[] toArrayWithoutWitnesses() {
+        try (ByteArrayOutputStream ms = new ByteArrayOutputStream()) {
+            try (BinaryWriter writer = new BinaryWriter(ms)) {
+                serializeWithoutWitnesses(writer);
+                writer.flush();
+                return ms.toByteArray();
+            }
+        } catch (IOException ex) {
+            throw new UnsupportedOperationException(ex);
```
csmuller: It would be nicer if some sort of SerializationException is thrown but this is how the `NeoSerializableInterface.toArray()` method does it. Introducing a checked `SerializationExceptions` would require try-catch blocks in many places.

csmuller:
@gsmachado Can you enable the `develop-3.x` on Codacy. It says that it "is not being analyzed" yet.

gsmachado:
@csmuller I added the branch to Codacy. :-)

Issue #35, opened 2019-05-20
Migrate from rxJava1 to rxJava2


gsmachado:
As a developer, I would like to get better support, performance, and maintainability by using rxJava2 lib. Also, rxJava2 is more compatible friendly for Android.

PR #104, opened 2019-12-08
rxjava2 and rx tests

### LINKED ISSUE #34: Tests for JsonRpc2_0Rx
As a developer, I would like to make really sure that the reactive interfaces (observables) are working as expected every time that something changes on the neow3j code.

gsmachado:
Even if the tests pass, we would need to test this code with a real Android application.
Also, I would like to also merge these changes to `develop-3.x`.

Resolves #34 and #35


core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
```diff
@@ -0,0 +1,296 @@
...
+            },
+            throwable -> fail(throwable.getMessage()),
+            () -> completedLatch.countDown());
+
+        transactionLatch.await(1, TimeUnit.SECONDS);
+        assertThat(results, equalTo(neoGetBlocks));
+
+        disposable.dispose();
+
+        completedLatch.await(1, TimeUnit.SECONDS);
```
csmuller: Why do we have to wait for the completedLatch here? Is the completion notification only emitted after the subscription (Disposable) is disposed? Is the observable not already complete after emitting all its blocks?

core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
```diff
@@ -0,0 +1,296 @@
...
+            expected.get(4),
+            expected.get(5),
+            expected.get(6)
+        );
+
+        OngoingStubbing<NeoGetBlock> stubbingNeoGetBlock =
+            when(neow3jService.send(any(Request.class), eq(NeoGetBlock.class)));
+
+        for (int i = 0; i < 7; i++) {
+            stubbingNeoGetBlock = stubbingNeoGetBlock.thenReturn(neoGetBlocks.get(i));
```
csmuller: Why is `neoGetBlocks` used here instead of `expected`?

core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
```diff
@@ -0,0 +1,296 @@
...
+        for (int i = 4; i < 7; i++) {
+            Thread.sleep(2000);
+            BigInteger added = BigInteger.valueOf(neoGetBlocks.get(i).getBlock().getIndex());
+            neoBlockCount.setResult(added);
+            stubbingNeoBlockCount = stubbingNeoBlockCount.thenReturn(neoBlockCount);
+        }
+
+        transactionLatch.await(15250, TimeUnit.MILLISECONDS);
+        assertThat(results.size(), equalTo(expected.size()));
+        assertThat(results, new ReflectionEquals(expected));
```
csmuller: Not sure what the ReflectionEquals does. Are we trying to match the two lists, `results` and `expected`, in any order?

core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
gsmachado: Addressed.

core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
gsmachado: addressed.

core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
gsmachado: Addressed.

core/src/test/java/io/neow3j/protocol/rx/JsonRpc2_0RxTest.java
```diff
@@ -0,0 +1,291 @@
...
+        assertThat(results, equalTo(expected));
+
+        disposable.dispose();
+
+        assertTrue(disposable.isDisposed());
+        assertThat(transactionLatch.getCount(), is(0L));
+    }
+
+    @Test
+//    @Ignore("Ignored due to a missing feature. "
```
csmuller: We forgot to uncomment the `@Ignore` again.

gsmachado:
Ok. We have problems with codacy due to code style, etc. I will merge this pull request since the feature + the CI-CD passed.

PR #110, opened 2020-03-11
Wrapper class for NEP5 standard support

mialbu:
This class provides neow3j with an abstraction of interacting with NEP5 token contracts.


contract/src/main/java/io/neow3j/contract/Nep5.java
```diff
@@ -0,0 +1,223 @@
...
+                .invoke();
+
+        return invocation.getResponse().getResult();
+    }
+
+    /**
+     * Builds a parameter list that can be used to build a ContractInvocation.
+     *
+     * @param from an account
+     * @param to a script hash of an address
```
gsmachado:
```suggestion
     * @param to the address, represented as a {@link ScriptHash}, that receives the tokens.
```

contract/src/main/java/io/neow3j/contract/Nep5.java
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Gets the symbol of the token.
+     *
+     * @return the symbol of the token.
+     * @throws IOException              if a connection problem with the RPC node arises.
+     * @throws ErrorResponseException   if the call to the node lead to an error. Not due to the
+     *                                  contract invocation itself but due to the call in general.
+     */
+    public String symbol() throws IOException, ErrorResponseException {
+        return this.testInvoke("symbol", null).getStack().get(0).asByteArray().getAsString();
```
gsmachado: We need to make this a bit more robust.
What if the `.get(0)` doesn't exist? It will break everything.
Can't we use `.stream().map()` here, for example? Or, `.stream().findFirst()`?

Example:
```
        return abi.getMethods()
            .stream()
            .filter(f -> !isEmpty(f.getName()))
            .filter(f -> f.getName().equals(functionName))
            .findFirst()
            .map(NeoContractMethod::getParameters)
            .orElse(Collections.emptyList());
```

Lamdas and streaming are your friends. :-)

Please, also change other lines that use `.get(0)` and make one test for it. Imagine that, somehow, for whatever reason, the node returns something wrong or not expected by us in the stack (i.e., imagine that position 0 does not exist).

contract/src/main/java/io/neow3j/contract/Nep5.java
```diff
@@ -0,0 +1,223 @@
...
+import io.neow3j.protocol.core.methods.response.InvocationResult;
+import io.neow3j.protocol.exceptions.ErrorResponseException;
+import io.neow3j.wallet.Account;
+
+import java.io.IOException;
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Nep5 {
```
gsmachado: Can you also add a brief Java Docs to the class level?
Very simple and short, like: "Wrapper to interact with contracts which are compatible with the NEP-5 standard."

gsmachado:
Good pull request. 😄 👍 
Minor changes are requested. Then, we can merge.

PR #119, opened 2020-04-22
Functionality for handling URI scheme

### LINKED ISSUE #115: Add NEP-9 support
Create a Java class that can generate URI's compatible with [NEP-9](https://github.com/neo-project/proposals/blob/master/nep-9.mediawiki).

mialbu:
Generate and read URI's compatible with [NEP-9](https://github.com/neo-project/proposals/blob/master/nep-9.mediawiki).
Closes #115


utils/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        public Builder amount(String amount) {
+            // TODO: 20.04.20 Michael: handle decimals if needed.
+            this.query.put("amount", amount);
+            return this;
+        }
+
+        public Builder amount(int amount) {
+            this.query.put("amount", String.valueOf((double) amount));
```
gsmachado, line 57: This conversion of types here is a bit dangerous.
Can you use `Integer` instead? Then you could do: `Integer.toString(amount);`

utils/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -0,0 +1,184 @@
...
+            this.address = address;
+            return this;
+        }
+
+        public Builder asset(String asset) {
+            this.query.put("asset", asset);
+            return this;
+        }
+
+        public Builder amount(String amount) {
```
gsmachado, line 50: Can you also add convenience methods like: `amount(BigDecimal amount)`, `amount(BigInteger amount)`, etc?

Check in our other classes how we're handling the amounts (related to types).

utils/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -0,0 +1,184 @@
...
+
+public class NeoURI {
+
+    private String uri;
+
+    protected NeoURI(Builder builder) {
+        this.uri = builder.uri;
+    }
+
+    public String getUri() {
```
gsmachado, line 19: can you also return a `java.net.URI`?

Then, we would have:

`public URI getUri()`

`public URI getUriAsString()`

utils/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -0,0 +1,184 @@
...
+
+import io.neow3j.utils.AddressUtils;
+import io.neow3j.utils.Strings;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class NeoURI {
```
gsmachado, line 11: For the parsing, I would create:

`public static NeoURI fromURI(String)`

or something like this.

You should search our code base for more examples.

utils/src/main/java/io/neow3j/contract/NeoURI.java
mialbu: Added both methods
(I replaced the URI string with a URI object that is generated in the builder.)

contract/src/main/java/io/neow3j/contract/Nep5Token.java
```diff
@@ -6,6 +6,8 @@
 import io.neow3j.wallet.Account;
 import io.neow3j.wallet.Wallet;
 import io.neow3j.wallet.exceptions.InsufficientFundsException;
+import org.bouncycastle.util.Strings;
```
gsmachado, line 9: this should not be here. :-)

contract/src/main/java/io/neow3j/contract/Nep5Token.java
```diff
@@ -161,4 +163,14 @@ public String transfer(Wallet wallet, ScriptHash to, BigDecimal amount)
                 .send();
     }
 
+    public String transferFromURI(Wallet wallet, NeoURI neoURI) throws IOException, ErrorResponseException {
+        ScriptHash to = neoURI.getAddress(); // TODO: 22.04.20 Michael: handle insufficient uri
+        BigDecimal amount = neoURI.getAmount();
+
+        if (this.scriptHash != neoURI.getAsset()) {
```
gsmachado, line 170: I don't get what's the logic here... can you explain a bit?

utils/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -35,6 +47,28 @@ public URI getUri() {
         return this.uri;
     }
 
+    public ScriptHash getAddress() {
+        return ScriptHash.fromAddress(this.address);
```
gsmachado, line 51: what if the provided address is not a valid Neo address? Exception? or how does it behave?

Also, do you have a test for it? :-)

utils/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -35,6 +47,28 @@ public URI getUri() {
...
+        String asset = this.query.get("asset");
+        return new ScriptHash(asset);
+    }
+
+    public BigDecimal getAmount() {
+        String amount = this.query.get("amount");
+        return new BigDecimal(amount);
+    }
+
+    public List<String> getHashs() {
```
gsmachado, line 64: it should be "getHashes"

utils/src/main/java/io/neow3j/contract/NeoURI.java
mialbu: `this.address` can only be set within the builder, and is checked with `AddressUtils.isValidAddress(this.address)` on line 192.

contract/src/main/java/io/neow3j/contract/Nep5Token.java
mialbu: idk how that was created, I'll remove it :)

contract/src/main/java/io/neow3j/contract/Nep5Token.java
mialbu: In the [Nep9](https://github.com/neo-project/proposals/blob/master/nep-9.mediawiki) proposal, a valid nep-5 smart contract script hash is required in the uri for smart contract invocations. This should be a check that the NeoURI's contract scripthash is equal to the Nep5Token instance.

gsmachado: In Java, a safe comparison of string is with:
`.equals()`
If that's the case, please change it. Just check, as well, if the string you want to compare is not null.

contract/src/main/java/io/neow3j/contract/Nep5Token.java
```diff
@@ -161,4 +162,14 @@ public String transfer(Wallet wallet, ScriptHash to, BigDecimal amount)
                 .send();
     }
 
+    public String transferFromURI(Wallet wallet, NeoURI neoURI) throws IOException, ErrorResponseException {
+        ScriptHash to = neoURI.getAddress(); // TODO: 22.04.20 Michael: handle insufficient uri
```
gsmachado, line 166: yes, here I would also check whether the address is null or not as well.

contract/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Creates a NeoURI from a NEP-9 URI String.
+     *
+     * @param uriString A NEP-9 URI String.
+     * @return A NeoURI object
+     * @throws IllegalFormatException if the provided URI has an invalid format.
+     */
+    public static NeoURI fromURI(String uriString) throws IllegalFormatException {
+        String[] base_and_query = uriString.split("\\?");
```
gsmachado, line 48: var names in cammel case

contract/src/test/java/io/neow3j/contract/Nep5TokenTest.java
```diff
@@ -79,6 +83,23 @@ public void transferGas() throws Exception {
                 is(w.getDefaultAccount().getVerificationScript()));
     }
 
+    @Ignore
+    @Test
+    public void transferTokenFromURI() throws IOException, ErrorResponseException {
+        setUpWireMockForSendRawTransaction();
+        setUpWireMockForInvokeFunction("transfer", "invokefunction_transfer.json");
+        setUpWireMockForInvokeFunction("decimals", "invokefunction_decimals.json");
+//        setUpWireMockForGetBlockCount();
```
gsmachado, line 92: can you remove it? 😄

mialbu: done :-)

contract/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -45,8 +45,8 @@ public NeoURI(Neow3j neow3j) {
      * @throws IllegalFormatException if the provided URI has an invalid format.
      */
     public static NeoURI fromURI(String uriString) throws IllegalFormatException {
-        String[] base_and_query = uriString.split("\\?");
-        String[] beginTx = base_and_query[0].split(":");
+        String[] baseAndQuery = uriString.split("\\?");
```
gsmachado, line 48: What if the `uriString` is null? NullPointerException? 😄 

Add a check for this, please. Also, a small unit test to make sure.

contract/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -45,8 +45,8 @@ public NeoURI(Neow3j neow3j) {
...
+        String[] beginTx = baseAndQuery[0].split(":");
```
gsmachado, line 49: Is there any chance that `baseAndQuery` does not have a `0` index position?

contract/src/main/java/io/neow3j/contract/NeoURI.java
```diff
@@ -45,8 +45,8 @@ public NeoURI(Neow3j neow3j) {
...
 
         if (beginTx.length != 2 || !beginTx[0].equals(NEO_SCHEME)) {
```
gsmachado, line 51: Same here: can we guarantee that `beginTx` will always have a `0` position?

contract/src/main/java/io/neow3j/contract/NeoURI.java
mialbu: Yes, I change that in the next commit. I now throw an IllegalArgumentException if uriString is null.

contract/src/main/java/io/neow3j/contract/NeoURI.java
mialbu: No, even if the uriString is an empty String or if the first entry of the uriString is the regex, the `.split()` will add an empty String to the index `0`. So such a uriString is then handled in the `if` condition below and a IllegalArgumentException is thrown.

contract/src/main/java/io/neow3j/contract/NeoURI.java
mialbu: Yes. Same reason as in the above conversation. :-)

gsmachado:
Great. I think we reached a reasonable state of this feature, indeed. I will commit and delete the source branch. 😄

PR #124, opened 2020-05-13
Feature/neo3 jsonrpc

mialbu:
As a user, I would like to call new Neo3 JSON-RPC methods. Also, I expect that old Neo2 methods are properly deprecated in the neow3j method API.


gsmachado:
@mialbu there are some things still missing, like tests for the nep5transfers with dates and stuff.... but I will merge this since we need the JSON-RPC functions to further develop on the `develop-3.x` branch.

gsmachado:
I will also delete the source branch.

PR #136, opened 2020-05-29
Fixes and tests for the Invocation.call() method

csmuller:


gsmachado:
hahaha, how is this possible? you added tests, but the test coverage decreased? 😄

contract/src/main/java/io/neow3j/contract/Invocation.java
```diff
@@ -279,16 +279,31 @@ public InvocationBuilder failOnFalse() {
...
+         * are required for a successful <tt>invokefunction</tt> call.
          *
          * @return the call's response.
          * @throws IOException if something goes wrong when communicating with the neo-node.
          */
         public NeoInvokeFunction call() throws IOException {
-            // This list is required for `invokescript` calls that will hit ChecekWitness checks
-            // in the smart contract.
-            String[] signers = getSigners();
+            // The list of signers is required for `invokefunction` calls that will hit
```
gsmachado, line 292: great! yes, it's better to add in this case, indeed...

gsmachado:
Oh. I've seen now that it's a draft. Sorry. Anyway, once more changes are pushed I can review again. 🙏

csmuller:
> Oh. I've seen now that it's a draft. Sorry. Anyway, once more changes are pushed I can review again. 🙏

Ok. I will increase the coverage and then push again.

csmuller:
Added some more unit tests on `Invocation` and `Nep5Token`.

contract/src/test/java/io/neow3j/contract/InvocationTest.java
```diff
@@ -35,12 +39,16 @@
...
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 public class InvocationTest {
 
     @Rule
     public WireMockRule wireMockRule = new WireMockRule();
 
+    @Rule
+    public ExpectedException exceptionRule = ExpectedException.none();
```
gsmachado, line 50: oh, I didn't know about this. Nice!

contract/src/test/java/io/neow3j/contract/Nep5TokenTest.java
```diff
@@ -131,4 +132,32 @@ public void getBalanceOfWallet() throws Exception {
         Nep5Token token = new Nep5Token(GasToken.SCRIPT_HASH, this.neow);
         assertThat(token.getBalanceOf(w), is(new BigInteger("411285799730")));
     }
+
+    @Test(expected = InsufficientFundsException.class)
```
gsmachado, line 136: hmmm, can you migrate this to the `@Rule` and you've done in the `InvocationTest`?

csmuller: I started to use the `ExpectedException` when I need more granular control over which function call throws the exceptino. In this example it is not necessary because that last function is the only one that will throw the `InsufficientFundsException`.

However, we can definitely change our style to always use `ExpectedException` instead of the annotation on the test method. I see no downside to it. Though, I wouldn't start to adapt it in existing unit tests, that's definitely not priority work.

gsmachado: Great! Clear now. 👍

gsmachado:
@csmuller just FYI, I will delete the source branch.

PR #137, opened 2020-06-01
Allow GAS balance checks before sending an invocation transaction

csmuller:


csmuller:
@gsmachado What do you think about this implementation. It doesn't force the dev to catch an exception that is automatically thrown if the sender's balance doesn't cover the GAS fees. The dev can check by himself by calling `canSenderCoverFees()`.

csmuller:
Got to ignore the red flag on the test coverage of `Invocation` in this pull-request. The coverage is improved in request #136.

gsmachado:
@csmuller 

What would be super nice to do, and we could use in our code more frequently, would be:

```
public Invocation canSenderCoverFeesOrExecute(Function<T, V> functionToExecuteIfCantCoverFees)
```

What do you think? Then the developer would define a function that would be executed -- not necessarily an exception... but maybe, just a `System.out.println()` or whatever it is defined.

contract/src/main/java/io/neow3j/contract/Invocation.java
```diff
@@ -156,6 +159,50 @@ public void addWitnesses(Witness... witnesses) {
...
+     * @return this.
+     * @throws IOException if something goes wrong in the communication with the neo-node.
+     */
+    public Invocation doIfSenderCannotCoverFees(BiConsumer<BigInteger, BigInteger> consumer)
+            throws IOException {
+        BigInteger fees = BigInteger.valueOf(
+                this.transaction.getSystemFee() + this.transaction.getNetworkFee());
+        BigInteger senderGasBalance = new GasToken(this.neow)
+                .getBalanceOf(this.transaction.getSender());
+        if (fees.compareTo(senderGasBalance) < 0) {
```
csmuller, line 176: Couldn't extract method from this because the fee and balance values are needed for the consumer below.

PR #140, opened 2020-06-08
Implementation of contract deployment for Neo 3

### LINKED ISSUE #101: Smart Contract deployment
Enable neow3j users to deploy smart contracts using neow3j.

- [x] Contract Manifest and ScriptHeader
- [x] Refactor of `ContractDeploy` class and related ones

csmuller:
Closes #101


contract/src/main/java/io/neow3j/contract/NefFile.java
```diff
@@ -0,0 +1,200 @@
...
+ * +------------+-----------+------------------------------------------------------------+
+ * | Checksum   | 4 bytes   | Sha256 of the header (CRC)                                 |
+ * +------------+-----------+------------------------------------------------------------+
+ * | Script     | Var bytes | Var bytes for the payload                                  |
+ * +------------+-----------+------------------------------------------------------------+
+ */
+public class NefFile extends NeoSerializable {
+
+    // NEO Executable Format 3 (NEF3)
+    private static final int MAGIC = 0x3346454E;
```
gsmachado, line 36: haha, it's the "magic". 😄

utils/src/main/java/io/neow3j/io/BinaryReader.java
```diff
@@ -274,7 +289,7 @@ public short readShort() throws IOException {
     }
 
     public byte[] readVarBytes() throws IOException {
-        return readVarBytes(0x7fffffc7);
+        return readVarBytes(0x1000000);
```
gsmachado, line 292: Was this wrong before? Or things changed for Neo3?

utils/src/main/java/io/neow3j/constants/NeoConstants.java
```diff
@@ -113,7 +113,7 @@
 
     //endregion
 
-    //region Transactions
+    //region Transactions & Contracts
```
gsmachado, line 116: Maybe I'm missing something (or my memory needs to be refreshed a bit), but why are we using these "region" annotations?

csmuller: Those are just there for giving a bit of structure. Intellij recognizes those and allow for collapsing.

contract/src/main/java/io/neow3j/contract/NefFile.java
csmuller: Neo core uses a magic number here and also when creating hashes e.g. for signature generation. For lack of a better name i went with the same.

utils/src/main/java/io/neow3j/io/BinaryReader.java
csmuller: Around 16MB (0x1000000) is the max used in neo-core, though, it is not a recent change in their code. It has been like that since 2018.

PR #150, opened 2020-06-24
Account and Wallet Refactor

### LINKED ISSUE #135: Consider removing the Builder from the Account class
`Account` allows instantiation only via a set of static creation methods. These methods return a builder that can be further configured with the properties label, isDefault, and isLocked.
The point of those creation methods is to make sure that only correctly configured accounts, i.e., with at least an address, can be created. 

Because these methods are static and return an `Account` or `Account.Builder` instance, it is impossible to extend `Account` and extend these methods. An extending subclass cannot reuse that code.

The properties label, isDefault, and isLocked - currently configurable only on the Builder - could be made available via setters without averse effects on the account state. Then the Builder becomes redundant. Furthermore, the static creation methods can then be converted into normal constructors because they can return an `Account` directly (instead of a `Builder`).

mialbu:
According to issue #135 the Account does not necessarily need the Builder pattern. 

In this pull request, the method `label` and `isLocked` of class `Account` can now be accessed via setters. The Wallet as well does not necessarily need the Builder pattern, which was also changed.
Additionally, Accounts now contain a parameter `wallet` that links the account to the wallet it is contained. This allows the `default` status to be handled within the Wallet. This also means that an Account now can only have a true default status, if it is linked to a wallet and it is this wallet's default account.

To add a new default Account to the wallet, the account has to be added and then set to default.

Resolves #135


mialbu:
The Wallet and Account are still used in several tests with the old Builder pattern. That is why the builds are failing. I will adapt those soon and then request a review.

wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -100,39 +92,91 @@ public ScryptParams getScryptParams() {
...
      *
      * @param account The account to add.
-     * @return true if the account was added, false if an account with the same address was already
-     * in the wallet.
+     * @return This wallet instance.
      */
-    public boolean addAccount(Account account) {
-        if (this.accounts.containsKey(account.getScriptHash())) {
-            return false;
+    public Wallet addAccount(Account account) {
```
csmuller, line 138: You can remove the addAccount(...) method. We normally only provide the plural version of such methods.

wallet/src/main/java/io/neow3j/wallet/Account.java
```diff
@@ -75,26 +73,54 @@ public String getLabel() {
...
+    public Account setLocked() {
+        this.isLocked = true;
+        return this;
+    }
+
+    public void unsetLocked() {
+        this.isLocked = false;
+    }
+
+    public void setWallet(Wallet wallet) {
```
csmuller, line 120: `setWallet()` should not be public but package-private or protected. The developer should not be able to set the wallet on an account, because this could lead to inconsistent states.

wallet/src/main/java/io/neow3j/wallet/Account.java
```diff
@@ -75,26 +73,54 @@ public String getLabel() {
...
+        if (this.wallet == null) return false;
+        return this.wallet.isDefault(this.getScriptHash());
     }
 
     // This method is required by the Wallet but must not be available to the developer because
     // it might bring a wallet into inconsistent state, i.e. having multiple default accounts.
-    void setDefault() {
-        this.isDefault = true;
-    }
+//    void setDefault() {
```
csmuller, line 97: Remove commented code.

wallet/src/main/java/io/neow3j/wallet/Account.java
```diff
@@ -75,26 +73,54 @@ public String getLabel() {
...
+//        this.isDefault = true;
+//    }
 
     // This method is required by the Wallet but must not be available to the developer because
     // it might bring a wallet into inconsistent state, i.e. having multiple default accounts.
-    void unsetDefault() {
-        this.isDefault = false;
-    }
+//    void unsetDefault() {
```
csmuller, line 103: Remove commented code.

wallet/src/main/java/io/neow3j/wallet/Account.java
```diff
@@ -245,64 +271,55 @@ public NEP6Account toNEP6Account() {
      *                           transactions.
      * @return the multi-sig account builder;
      */
-    public static Builder fromMultiSigKeys(List<ECPublicKey> publicKeys, int signatureThreshold) {
```
csmuller, line 248: This comment has nothing to do with your changes. I just realized that this method's name is not intuitive. Let's rename it to `createMultiSigAccount()`.

wallet/src/main/java/io/neow3j/wallet/Account.java
```diff
@@ -37,21 +38,18 @@
     private String address;
     private String encryptedPrivateKey;
     private String label;
-    private boolean isDefault;
     private boolean isLocked;
     private VerificationScript verificationScript;
+    private Wallet wallet;
 
-    private Account() {
+    public Account() {
```
csmuller, line 45: Make default constructor private.

wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -142,6 +186,11 @@ public boolean addAccount(Account account) {
      * @return true if an account was removed, false if no account with the given address was found.
      */
     public boolean removeAccount(ScriptHash scriptHash) {
+        if (scriptHash == this.getDefaultAccount().getScriptHash()
```
csmuller, line 189: Use `.equals()` instead of `==`. The `scriptHash` argument might not be the same object that is referenced in the default account. In such a scenario the boolean expression would evaluate to false, but should.

wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -142,6 +186,11 @@ public boolean addAccount(Account account) {
...
+                && this.accounts.size() > 1) {
+            ScriptHash newDefaultAccount = this.accounts.entrySet().iterator().next().getKey();
+            this.setDefaultAccount(this.accounts.get(newDefaultAccount).getScriptHash());
+        }
         return accounts.remove(scriptHash) != null;
```
csmuller, line 194: - Need to remove the account's reference to the wallet from the account (i.e. set to null).
- Need to clear the wallet's `defaultAccount` variable if the last account is removed (i.e. set to null)
- Extend corresponding unit test to cover this behavior.

wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -166,35 +215,35 @@ public NEP6Wallet toNEP6Wallet() {
...
 
-    public static Builder fromNEP6Wallet(InputStream nep6WalletFileInputStream) throws IOException {
+    public static Wallet fromNEP6Wallet(InputStream nep6WalletFileInputStream) throws IOException {
         NEP6Wallet nep6Wallet = OBJECT_MAPPER
                 .readValue(nep6WalletFileInputStream, NEP6Wallet.class);
         return fromNEP6Wallet(nep6Wallet);
     }
 
-    public static Builder fromNEP6Wallet(NEP6Wallet nep6Wallet) {
+    public static Wallet fromNEP6Wallet(NEP6Wallet nep6Wallet) {
```
csmuller, line 237: A NEP6Wallet will contain a account marked as default. I.e. you need to look for the default account in the NEP6Wallet and set it on the Wallet.

wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -100,39 +92,91 @@ public ScryptParams getScryptParams() {
...
+        return this.defaultAccount != null;
+    }
+
+    /**
+     * Checks whether an account is the default account in the wallet.
+     *
+     * @param accountScriptHash the account to be checked.
+     * @return Whether the given account is the default account in this wallet.
+     */
+    public Boolean isDefault(ScriptHash accountScriptHash) {
```
csmuller, line 112: Use `.equals()` instead of `==`.

wallet/src/main/java/io/neow3j/wallet/Wallet.java
```diff
@@ -42,20 +39,16 @@
     private String version;
     private Map<ScriptHash, Account> accounts = new HashMap<>();
     private ScryptParams scryptParams;
+    private ScriptHash defaultAccount;
 
     static {
         addBouncyCastle();
     }
 
-    private Wallet() {
```
csmuller, line 50: Add Javadoc mentioning that the `Wallet` will be empty (no accounts).

wallet/src/test/java/io/neow3j/wallet/WalletTest.java
```diff
@@ -93,32 +92,24 @@ public void addAccount() throws InvalidAlgorithmParameterException,
...
-        Account acct = Account.fromECKeyPair(ECKeyPair.createEcKeyPair()).build();
-        assertTrue(w.addAccount(acct));
-        assertFalse(w.addAccount(acct));
+        Account acct = new Account(ECKeyPair.createEcKeyPair());
+
+        // Add account to wallet.
+        w.addAccount(acct);
+        assertTrue(w.getAccounts().contains(acct));
+        // Adding an account twice does not change the wallet.
+        assertThat(w.addAccount(acct), is(w));
```
csmuller, line 112: This assertion is always true because the returned wallet is always the same as the one the method is called on. Only thing we could assert here is if the accounts list still contains only one account.

mialbu: I changed this, to compare the number of accounts in the wallet.

wallet/src/test/java/io/neow3j/wallet/WalletTest.java
```diff
@@ -120,13 +125,28 @@ public void testRemoveAccounts() throws InvalidAlgorithmParameterException,
...
+    @Test(expected = NullPointerException.class)
+    public void testRemoveAccounts_lastRemainingAccount() {
+
+        final String address = "AUcY65mkxygUB5bXZqYhNKsrq1khuncqr3";
+        Wallet w = Wallet.createWallet();
+        Account lastRemainingAccount = w.getAccounts().get(0);
+
+        assertEquals(w, lastRemainingAccount.getWallet());
+        assertEquals(lastRemainingAccount, w.getDefaultAccount());
+        assertTrue(w.removeAccount(lastRemainingAccount.getScriptHash()));
```
csmuller, line 144: A NullPointerException is thrown here, but shouldn't. The test only succeeds because it expects a NullPointerException in the annotation, but shouldn't.

The last statement in this test should be a assertion like `assertThat(w.getDefaultAccount(), nullValue())`

mialbu: I changed the structure of the Wallet a bit considering this.
Now, the empty constructor is private and a Wallet can only be initialised containing at least one account.
A wallet without any accounts is no longer possible. If one wants to remove the last account in the wallet, an exception is thrown. Further, in a wallet, there is always one account that is the default account. If the default account is removed, the list of accounts in the wallet is iterated and the next account is set as default.

mialbu:
I adapted the sign method of the Invocation class according to the changes in the wallet. It now iterates through the cosigners using `foreach()` and uses the new method `holdsAccount()` from the wallet.

gsmachado:
Let's wait for @csmuller feeedback. 👍 Because I haven't checked in very detailed, though.

PR #152, opened 2020-06-30
Refactor `StackItem` to use Jackson annotations

### LINKED ISSUE #131: Refactor `StackItem` to use Jackson annotations for deserializing subclasses
The `StackItem` class currently uses a custom deserializer to deserialize its various subclasses. 
See `TransactionAttribute` as an example on how to do this.

mialbu:
According to issue #131,  StackItems could potentially be deserialised using Jackson annotations.

Resolves #131

### DIFF
--- a/core/src/main/java/io/neow3j/protocol/core/methods/response/InteropInterfaceStackItem.java
+++ b/core/src/main/java/io/neow3j/protocol/core/methods/response/InteropInterfaceStackItem.java
@@ -1,0 +1,52 @@
+package io.neow3j.protocol.core.methods.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.neow3j.contract.ScriptHash;
+import io.neow3j.model.types.StackItemType;
+import io.neow3j.utils.BigIntegers;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.math.BigInteger;
+import java.util.Arrays;
+import java.util.Objects;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class InteropInterfaceStackItem extends StackItem {
+
+    @JsonProperty("value")
+    private String value;
+
+    public InteropInterfaceStackItem() {
+        super(StackItemType.INTEROP_INTERFACE);
+    }
+
+    public InteropInterfaceStackItem(String value) {
+        super(StackItemType.INTEROP_INTERFACE);
+        this.value = value;
+    }
+
+    /**
+     * Decodes the stack item's value and returns it.
+     *
+     * @return the value of this stack item.
+     */
+    public String getValue() {
+        return this.value;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof io.neow3j.protocol.core.methods.response.InteropInterfaceStackItem)) return false;
+        io.neow3j.protocol.core.methods.response.InteropInterfaceStackItem other =
+                (io.neow3j.protocol.core.methods.response.InteropInterfaceStackItem) o;
+        return getType() == other.getType()
+                && getValue().equals(other.getValue());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getValue());
+    }
+}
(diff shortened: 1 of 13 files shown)


core/src/main/java/io/neow3j/protocol/core/methods/response/InteropInterfaceStackItem.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public String getValue() {
+        return this.value;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof io.neow3j.protocol.core.methods.response.InteropInterfaceStackItem)) return false;
+        io.neow3j.protocol.core.methods.response.InteropInterfaceStackItem other =
```
gsmachado, line 42: strange.... why using the fully qualified class name?

model/src/main/java/io/neow3j/model/types/StackItemType.java
```diff
@@ -15,6 +15,17 @@
     MAP("Map", 0x48),
     INTEROP_INTERFACE("InteropInterface", 0x60);
 
+    public static final String ANY_VALUE = "Any";
```
gsmachado, line 18: hmmmmmm...... isn't it possible to use the Enums? If not, then, use the values from below to the ENUMs value.... like this:

```
ANY(ANY_VALUE, 0x00);
```

If you can't do this, then declare `ANY_VALUE` as `public static final String` somewhere else (like in a file called `StackItemTypeConstants.java`.

core/src/test/java/io/neow3j/protocol/core/methods/response/StackItemTest.java
```diff
@@ -115,6 +115,7 @@ public void testDeserializeIntegerStackItem() throws IOException {
         IntegerStackItem item = rawItem.asInteger();
         assertEquals(new BigInteger("1124"), item.getValue());
 
+        // TODO: 23.06.20 Michael: set empty content to default value - here to BigInteger.ZERO
```
gsmachado, line 118: yes, this is important, IMHO!

mialbu: Actually, I've already implemented this with the following setter in `IntegerStackItem.java`. I'll remove the todo.

```java
    @JsonSetter("value")
    public void setValue(BigInteger value) {
        if (value != null) {
            this.value = value;
        } else {
            this.value = BigInteger.ZERO;
        }
    }
```

model/src/main/java/io/neow3j/model/types/StackItemType.java
mialbu: It's not possible to use the enums, since the Subtypes' name declaration in the abstract class StackItem must be final. I changed it according your suggestion.

core/src/main/java/io/neow3j/protocol/core/methods/response/InteropInterfaceStackItem.java
mialbu: Oh, I didn't notice this. Idk, my IDE must have filled it automatically. I'll change this.

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
```diff
@@ -1,32 +1,38 @@
...
-import java.util.Map;
-import java.util.Objects;
-
-@JsonDeserialize(using = StackDeserializer.class)
-public class StackItem {
 
+@JsonTypeInfo(use = Id.NAME, property = "type", include = As.EXISTING_PROPERTY)
+@JsonSubTypes(value = {
+        @JsonSubTypes.Type(value = AnyStackItem.class, name = StackItemType.ANY_VALUE),
+//        @JsonSubTypes.Type(value = PointerStackItem.class, name = StackItemType.POINTER_VALUE),
```
gsmachado, line 15: Could you add a Todo comment for this? Then we could easily find what's missing in the future.

core/src/main/java/io/neow3j/protocol/core/methods/response/StackItem.java
```diff
@@ -1,32 +1,38 @@
...
+        @JsonSubTypes.Type(value = BooleanStackItem.class, name = StackItemType.BOOLEAN_VALUE),
+        @JsonSubTypes.Type(value = IntegerStackItem.class, name = StackItemType.INTEGER_VALUE),
+        @JsonSubTypes.Type(value = ByteStringStackItem.class, name = StackItemType.BYTE_STRING_VALUE),
+//        @JsonSubTypes.Type(value = BufferStackItem.class, name = StackItemType.BUFFER_VALUE),
```
gsmachado, line 19: same here (add Todo)

gsmachado:
Merged. I'm deleting the source branch. 😃

PR #157, opened 2020-07-13
Policy Contract Support

### LINKED ISSUE #117: Add PolicyContract support
Create a Java class implementing the "functions"/interface of PolicyContract.

Description:
https://github.com/neo-ngd/NEO3-Development-Guide/tree/master/en/SmartContract#policycontract

mialbu:
This pull request adds support to the native policy contract.

Closes #117.


contract/src/main/java/io/neow3j/contract/PolicyContract.java
```diff
@@ -0,0 +1,186 @@
...
+import io.neow3j.wallet.Wallet;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a Policy contract and provides methods to invoke it.
+ */
+public class PolicyContract extends SmartContract{
```
gsmachado, line 18: Minor: space after `SmartContract`

```suggestion
public class PolicyContract extends SmartContract {
```

contract/src/test/java/io/neow3j/contract/NeoTokenTest.java
```diff
@@ -88,7 +88,6 @@ public void getUnclaimedGas() throws IOException {
 
     @Test
     public void registerCandidate() throws IOException {
-        NeoConfig.setMagicNumber(new byte[]{0x01, 0x03, 0x00, 0x0}); // Magic number 769
```
gsmachado, line 91: Why was this removed?

contract/src/main/java/io/neow3j/contract/SmartContract.java
```diff
@@ -4,6 +4,7 @@
 import io.neow3j.constants.InteropServiceCode;
 import io.neow3j.constants.NeoConstants;
 import io.neow3j.contract.exceptions.UnexpectedReturnTypeException;
+import io.neow3j.crypto.ECKeyPair;
```
gsmachado, line 7: Are these imports in the `SmartContract` required? Sorry, I haven't opened the pull request in my IntelliJ to check that out.

contract/src/test/java/io/neow3j/contract/PolicyContractTest.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final ScriptHash CONSENSUS_MULTISIG_SCRIPT_HASH =
+            new ScriptHash("55b842d631f43f23257a27992ac2b53169a4fe00");
+    private static final String CONSENSUS_MULTISIG_ADDRESS = "AFs8hMHrS8emaPP4oyTuf5uKPuAW6HZ2DF";
+    private static final byte[] VERIFICATION_SCRIPT =
+            Numeric.hexStringToByteArray("110c2102c0b60c995bc092e866f15a37c176bb59b7ebacf069ba94c0ebf561cb8f956238110b41c330181e");
+
+    @Before
+    public void setUp() {
+        // Configuring WireMock to use default host and port "localhost:8080".
+        WireMock.configure();
```
gsmachado, line 45: Just a side question (maybe @csmuller can also help here): can we configure `WireMock` to use a random port each time (for each test suite)? Then, we could fetch the random port and inject it to the `HttpService` for each time, instead of using a static/fixed one (8080). This way we would prevent sporadic failures to tests... sometimes I see that the WireMock port is already in use, somehow...

contract/src/test/java/io/neow3j/contract/PolicyContractTest.java
```diff
@@ -0,0 +1,172 @@
...
+        PolicyContract policyContract = new PolicyContract(this.neow3j);
+        List<ScriptHash> blockedAccounts = policyContract.getBlockedAccounts();
+
+        assertNotNull(blockedAccounts);
+        assertThat(blockedAccounts.size(), is(2));
+        assertThat(blockedAccounts, contains(
+                ScriptHash.fromAddress("Aa1rZbE1k8fXTwzaxxsPRtJYPwhDQjWRFZ"),
+                ScriptHash.fromAddress("ATpVyfpFwE2SzNGSvXDNrtRyfVLajhn7yN")));
+    }
+
```
gsmachado, line 97: remove this line space/break, here.

utils/src/main/java/io/neow3j/constants/InteropServiceCode.java
```diff
@@ -45,7 +45,8 @@
 
     // Native tokens.
     NEO_NATIVE_TOKENS_NEO("Neo.Native.Tokens.NEO", null),
-    NEO_NATIVE_TOKENS_GAS("Neo.Native.Tokens.GAS", null);
+    NEO_NATIVE_TOKENS_GAS("Neo.Native.Tokens.GAS", null),
+    NEO_NATIVE_POLICY("Neo.Native.Policy", null); // TODO: 24.06.20 Michael: Is price null?
```
gsmachado, line 49: hmmm.. what do you mean by "is price null"? 😄 
like, the paid gas?

gsmachado: @csmuller just a clarification: so... it means that when the "gas price" is `null` in this file, there's some kind of dynamic calculation happening somewhere?

contract/src/main/java/io/neow3j/contract/SmartContract.java
mialbu: You're right, I'll remove it.

contract/src/test/java/io/neow3j/contract/NeoTokenTest.java
csmuller: We looked at this test as an example for setting the magic number and realized that specifically in this test it isn't even necessary.

utils/src/main/java/io/neow3j/constants/InteropServiceCode.java
mialbu: > hmmm.. what do you mean by "is price null"? 😄
> like, the paid gas?

Yes, the execution GAS price. @csmuller just told me, it actually should be set to null 👍 , so I'll remove the todo.

> @csmuller just a clarification: so... it means that when the "gas price" is `null` in this file, there's some kind of dynamic calculation happening somewhere?

Correct.

contract/src/test/java/io/neow3j/contract/NeoTokenTest.java
mialbu: In this test this is not necessary, I looked into it with @csmuller.

contract/src/test/java/io/neow3j/contract/PolicyContractTest.java
mialbu: I configured `WireMock`, such that it uses a dynamic port and then inject it to the `HttpService`. Have a look at it and after reviewing it, we could add that configuration also to other test classes that use `WireMock`.

Issue #168, opened 2020-07-27
Analyze and adapt changes in Neo3-preview3


csmuller:
Analyze the changes made in Neo3-preview3 when it is realeased (presumably in the beginning of August 2020), create a list of tasks required to get neow3j SDK up to date and implement the changes.

Changelist:
- [x] InteropServiceCode names (e.g. `Neo_Crypto_VerifyWithECDsaSecp256r1` and `Neo_Crypto_VerifyWithECDsaSecp256k1`)
- [x] Change `ContractParameterType`s and add `Any` and `Map`.
- Add following opcodes:
  - [x] TRY
  - [x] TRY_L
  - [x] ENDTRY
  - [x] ENDTRY_L
  - [x] ENDFINALLY
- Rename the following:
  - [x] `Cosigner` to `Signer`
  - [x] `StandbyValidators` to `StandbyCommittee`
  - [x] `gas_consumed` to `gasConsumed`
  - [x] `tcp_port` to `tcpPort`
  - [x] `ws_port` to `wsPort`
  - [x] `user_agent` to `userAgent`
  - [x] `sys_fee` to `sysfee`
  - [x] `net_fee` to `netfee`
  - [x] `valid_until_block` to `validuntilblock`
  - [x] `asset_hash` to `assethash`
  - [x] `last_updated_block` to `lastupdatedblock`
  - [x] `transfer_address` to `transferaddress`
  - [x] `block_index` to `blockindex`
  - [x] `transfer_notify_index` to `transfernotifyindex`
  - [x] `tx_hash` to `txhash`
  - [x] `vm_state` to `vmstate`
- [x] Adapt unit tests
- [x] Adapt integration tests
- [x] Adapt checksum calculation of NefFile (see merge-request #190)
- [x] Adapt `WitnessScope` (codes changed and one new scope was introduced)
- [x] Adapt Signer object

PR #190, opened 2020-08-11
Update NefFile implementation to Neo3-preview3

csmuller:
Solves part of #168. 
`NefFile` now uses SHA256 twice for checksum calculation.
I created a separate method for this on the `Hash` class which is now used in all the places that apply SHA256 twice.


gsmachado:
@csmuller so far, the PR seems fine. As you mentioned, once you fix the tests, I can take a look again on it.

contract/src/test/java/io/neow3j/contract/SmartContractTest.java
```diff
@@ -244,21 +250,27 @@ public void callFunctionWithParams() throws IOException {
                 is(BigInteger.valueOf(3000000000000000L)));
     }
 
+    @Ignore("The test fails because the mocked `incokescript` RPC call expects the script hash of "
+            + "the sender in as a parameter in the list of signers. But because of changed "
+            + "InteropServiceCalls, neow3j currently produces another script hash from the sender"
+            + " account as expected in this test. As soon as these changes to the "
+            + "InteropServiceCode are introduced to neow3j this test can be normally run.")
```
gsmachado, line 257: We just need to remember to remove the ignored of this test when @mialbu completes the adaptions.
Should we add a TODO here? No worries if not.

gsmachado:
It's fine for me. I just scrolled the code a bit to check inconsistencies and stuff. But let's merge this one and move on with dev. 👍 😄 💯

PR #193, opened 2020-08-24
MVP: gradle plugin

gsmachado:
The first version of the Gradle plugin. MVP state.


gradle-plugin/src/main/java/io/neow3j/devpack/gradle/Neow3jCompileAction.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.List;
+import org.gradle.api.Action;
+
+public class Neow3jCompileAction implements Action<Neow3jCompileTask> {
+
+    @Override
+    public void execute(Neow3jCompileTask neow3jPluginCompile) {
+        String canonicalClassName = neow3jPluginCompile.getOptions().getClassName();
+
+        System.out.println("Smart Contract class: "
```
csmuller, line 26: Not sure how thorough i should do this review, but in the end this print line should be removed.

gradle-plugin/src/main/java/io/neow3j/devpack/gradle/Neow3jCompileAction.java
```diff
@@ -0,0 +1,73 @@
...
+                + canonicalClassName);
+
+        ofNullable(canonicalClassName)
+                .orElseThrow(() ->
+                        new IllegalArgumentException("The parameter '"
```
csmuller, line 31: I like error messages ;) therefore I'd add a bit more info, like:
"The parameter Neow3jPluginOptions.CLASSNAME_NAME needs to be set in the Neow3jPluginTask.NEOW3J_COMPILER_OPTIONS_NAME options."

gradle-plugin/src/main/java/io/neow3j/devpack/gradle/Neow3jPluginUtils.java
```diff
@@ -0,0 +1,77 @@
...
+            });
+        });
+        return urls;
+    }
+
+    static String getClassName(String fqClassName) {
+        if (fqClassName == null) {
+            return null;
+        }
+        int firstChar;
```
csmuller, line 57: variable declaration and initialization can be on the same line.

csmuller (approved):
Good stuff!

build.gradle
```diff
@@ -33,6 +33,7 @@ buildscript {
 plugins {
     id 'com.jfrog.bintray' version '1.7.3'
     id 'org.unbroken-dome.test-sets' version '1.4.2'
+    id 'com.gradle.plugin-publish' version '0.12.0' apply false
```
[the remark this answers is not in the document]
gsmachado: Unfortunately, I didn't go deep on this one and I'd rather prefer to leave like this for the moment.

When I was trying to come up with a nice sub-module structure, I couldn't make it work with the `plugins` declaration clause within the `gradle-plugin/build.gradle`. I didn't go further on the analysis, so I kept consistency with what we had already (which was working). But your logic is correct: it would be nice to have project-specific things inside each module's `build.gradle` files, indeed.....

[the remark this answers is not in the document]
gsmachado: I will resolve this conversation for now.

Issue #167, opened 2020-07-27
Support loops


csmuller:
The compiler should support Java loops that can be mapped into NeoVM code. This includes `for`, `while`, and `do while` loops.
Keywords: `for`, `do`, `while`, `break`, `continue`
Part of this issue is determining if enhanced\* for loops can be compiled to NeoVM code too. But if the actual implementation needs a separate task depends on the determined implementation effort. 

\* An enhanced for loop is a construct like the following: `for (Entry<K,V> : someMap) { ... }`.

Issue #166, opened 2020-07-27
Support logical operators


csmuller:
The compiler should support all Java logical operators that can be mapped to NeoVM code.
Operators: `&&, ||, !`

Issue #165, opened 2020-07-27
Support relational operators


csmuller:
The compiler should support all Java relational operators that can be mapped to NeoVM code.
Operators: `>,   <, >=, <=, ==, !=`

Issue #164, opened 2020-07-27
Support bitwise operators


csmuller:
The compiler should support all Java bitwise operators that can be mapped to NeoVM code.
Operators: `&, |, ^, ~, <<,   >>, >>>`
Assignment operators: `&=, |=, ^=, <<=, >>=`

csmuller:
The logical right-shift operator `>>>` is not supported by NeoVM. Thus, it will not be supported by the neow3j-compiler.

gsmachado:
@csmuller should we close this issue, then?

csmuller:
> @csmuller should we close this issue, then?

No, all other bitwise operators (e.g. `>>` or `&`) must still be supported.

gsmachado:
Oh, yes. Good. 😄  👍

PR #196, opened 2020-08-26
Compiler features for Milestone 3.2.0

### LINKED ISSUE #161: Support conditional expressions
The compiler must support conditional expressions. 
Keywords: `if`, `else`, `else if`, `switch`, `case`, `default`, `break`
Operators: `?`, `:`, i.e. the ternary operator.

Requires:
- Mapping all conditional expressions from JVM to NeoVM.
- Dealing with jump addresses on JMP opcodes that point to an address in the same method.
- Mapping switch-cases to if-else code because NeoVM doesn't support switch-case.
- Clarifying if it is possible to map string-based swtich-cases to NeoVM code.

csmuller:
Resolves #161, #163, #164, #165, #166, #167.


gsmachado:
I'm fine with this. Should we merge this before #188 ?

gsmachado (approved):

Issue #199, opened 2020-08-31
Fixes, improvements, refactorings for the compiler and devpack


csmuller:
This issue is a collection of bugs and improvements found along the way that should be fixed and applied to the neow3j compiler and devpack.

Included in release 3.2.1:
- [x] Correct byte array instantiation. Simply use the `NEWBUFFER` NeoVM opcode instead of `NEWARRAY` opcode, when a byte array is instantiated. Optimizations for array initializations will be handled in #177 (in release 3.2.1)
- [x] Fix `Helper.toInt()` method and add missing `Helper.asByte()` method to devpack (in release 3.2.1)
- [x] Relational operators for integers work in reverse (in release 3.2.1)
- [x] Accessing a field variable doesn't use the correct owner class of the method (in release 3.2.1)
- [x] Smart contracts without any class-level annotations don't compile because the list of annotation is null (in release 3.2.1)

Included in release 3.3.0:
- [x] The `buildManifestExtra()` method only checks for the `@ManifestExtras` annotation and therefore misses the cases in which the developer only adds one `@ManifestExtra` to the contract class. Resolved in #215. 
- [x] The JVM instruction `DUP2` is not properly converted into NeoVM code, i.e. the produced NeoVM script can be erroneous. Resolved in #218.
- [x] Support the helper methods `addressToScriptHash()`, `hexToBytes()`, `stringToInt()` but without implementing a VM like the `CctorSubVM` from neo-devpack-dotnet. I.e., only support constant value parameters. The methods have been implemented and the compiler extended in #217.

csmuller:
Opened a new issue with the same intent like this one but for milestone 3.4.0. Like this we always know what was done in which milestone, which makes reporting easier.

PR #201, opened 2020-09-01
Fixes for byte array initialization and devpack helper methods

csmuller:
Part of #199

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/Compiler.java
+++ b/compiler/src/main/java/io/neow3j/compiler/Compiler.java
@@ -407,7 +407,12 @@ public class Compiler {
             // region ### ARRAYS ###
             case NEWARRAY:
             case ANEWARRAY:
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/io/neow3j/compiler/Compiler.java
```diff
@@ -407,7 +407,12 @@ private AbstractInsnNode handleInsn(NeoMethod neoMethod, AbstractInsnNode insn)
             // region ### ARRAYS ###
             case NEWARRAY:
             case ANEWARRAY:
-                neoMethod.addInstruction(new NeoInstruction(OpCode.NEWARRAY));
+                if (((IntInsnNode)insn).operand == 8) {
```
gsmachado, line 410: I think it's fine like this, for the moment. 👍  💯

gsmachado: If you feel like enhancing a bit the code, you could create more descriptive methods, like this: .isByteArray(insn)

PR #204, opened 2020-09-04
Small changes to the SmartContract API to make development easier

csmuller:
I need these changes to work with the `SmartContract` class in the compiler and I also think that they will be useful for other developers as well.


gsmachado:
@csmuller we mostly did this together. So, it's fine! I will merge now and let's move on. 😄 👍 💯

gsmachado (approved):

Issue #179, opened 2020-07-28
Create a Gradle plugin to compile Java contracts


csmuller:
The Gradle plugin can be used inside of a Gradle project and should provide something like a `compileNeo` task that compiles the smart contract contained in the Gradle project to NeoVM code.

gsmachado:
Still missing the following things:

- [ ] Output the ABI file and Manifest file (in the same output directory as the .nef)
- [ ] Multiple files support (i.e., multiple Smart Contracts)
- [ ] Test with a Kotlin project
- [ ] Documentation (how to use it, examples, etc)

gsmachado:
> * [x]  Output the ABI file and Manifest file (in the same output directory as the .nef)

So... for the moment, we're outputting the manifest, which includes the ABI.

Do we need to output the ABI as a separate file?

csmuller:
Not in my opinion. I don't see a reason to have it written to a files separately. The neon compiler does it, but it doesn't have more information than what's in the manifest.

gsmachado:
> Not in my opinion. I don't see a reason to have it written to a files separately. The neon compiler does it, but it doesn't have more information than what's in the manifest.

I think that @mialbu can help us here. 
The question is: does the Neo Blockchain Toolkit expects an ABI file or a Manifest file to deploy a Smart Contract?

If it requires a separate ABI file, then we do it. Otherwise, we just output the manifest (as currently is).

csmuller:
@gsmachado Is there anything left to do here for Milestone 3.4.0?

From the above todo list the item "Multiple files support" cannot be achieved because the compiler doesn't support multiple files anyway at the moment.
The documentation at https://neow3j.io/#/neo3_guides/compiler_devpack/compilation?id=gradle-plugin covers the "Documentation" todo.
The only thing that remains is "Test with a Kotlin project". I'm not sure what is meant by that. The contract to compile is in Java anyway, because the compiler is not Kotlin-compatible yet. Maybe you mean, using a Gradle config file written in Kotlin.

gsmachado:
> The only thing that remains is "Test with a Kotlin project". I'm not sure what is meant by that. The contract to compile is in Java anyway, because the compiler is not Kotlin-compatible yet. Maybe you mean, using a Gradle config file written in Kotlin.

What I meant by that is: we need to come up with a Kotlin project, and use the gradle plugin with it. Does it add the "neow3jCompile" gradle task? Does it work?

csmuller:
[...]

PR #205, opened 2020-09-04
#179: generate a manifest file

gsmachado:


gsmachado:
@csmuller this is the first version.

We still need to test this with the compiler, integrated.

csmuller (approved):

PR #206, opened 2020-09-04
change Gradle structure

gsmachado:


gsmachado:
@csmuller this is pretty critical for us.

We need to check whether we can properly release to jfrog + maven central once this is merged.

Maybe a strategy would be to first merge `develop-3.x` into this branch?

gsmachado:
@csmuller I will merge this.

It will not affect neow3j's functionality as I tested it. However, we might face some problems with Jacoco (codecov.io) and the overall test reports.

We can plan to fix that in another issue/PR.

PR #218, opened 2020-09-18
Fix conversion of JVM's special DUP and POP opcodes.

csmuller:
This fixes one point of issue #199.

I can't write tests for all the new code that I added, e.g., for the handling DUP_X2 JVM opcode. I simply don't know which Java code will result in such opcodes.


gsmachado (approved):
@csmuller I liked the comments inline with the code. 👍 

All fine for me. I will proceed and merge it.

Issue #194, opened 2020-08-25
Complete the implementation of all existing StackItem types


csmuller:
For the `StackItem` types `Buffer` and `Pointer` the neow3j SDK is lacking the DTO object to read from and write to JSON representation. This is needed, e.g., when receiving stack results from a script invocation.
The missing `StackItem`s need to be added.

- [x] Buffer StackItem Implementation
- [x] Pointer StackItem Implementation

mialbu:
This was merged in pr #220. I'll close this.

PR #220, opened 2020-09-28
Buffer and Pointer StackItem Implementation

mialbu:
According to #194, in this PR the Buffer and Pointer stack items are implemented.

### DIFF
--- a/core/src/main/java/io/neow3j/protocol/core/methods/response/BufferStackItem.java
+++ b/core/src/main/java/io/neow3j/protocol/core/methods/response/BufferStackItem.java
@@ -1,0 +1,85 @@
+package io.neow3j.protocol.core.methods.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/neow3j/protocol/core/methods/response/BufferStackItem.java
```diff
@@ -0,0 +1,85 @@
...
+//    /**
+//     * <p>Gets this byte array's value as an address.</p>
+//     * <br>
+//     * <p>Expects the byte array to be a script hash in little-endian order.</p>
+//     *
+//     * @return the address represented by this byte array.
+//     */
+//    public String getAsAddress() {
+//        return new ScriptHash(getValue()).toAddress();
+//    }
```
mialbu, line 47: @gsmachado Here I am unsure, whether this should be implemented in the buffer stack item. Could you give me some input here?

gsmachado: @mialbu as discussed, leave the `getAsAddress()` method. 👍

gsmachado (approved):

PR #221, opened 2020-09-28
Save NEP6 Wallet with custom set name

### LINKED ISSUE #189: Set the wallet name by their public key
The user Evixis in Discord #java gave the suggestion:

> it would be really awesome if WalletUtils.generateNewWalletFile() would name wallets by their public key and not "neow3j". Like web3j does, because it's easier to locate the wallet in that way.

Therefore:
- [x] leave the option to the user how the wallet name should be set
- [x] ~set the default name as the public key (encoded)~

mialbu:
closes #189


gsmachado:
@mialbu yes, as we discussed, I think this is the minimum we can do to satisfy users of neow3j on how to name the wallet, etc. I will approve this PR. 💯 👍

PR #230, opened 2020-10-07
Minor changes and an other refactoring for the compiler

csmuller:
This refactoring was about the classes `NeoModule`, `NeoInstruction`, `NeoMethod`, `NeoVariable`, and `NeoJumpInstruction`. I made their field variables private and therefore needed to introduce getters and setters. Also i moved all `Converter` classes to their own package to have a bit more order in the compiler package.

Besides the refactoring I added a few more compiler tests and the utils methods that @gsmachado proposed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ The neow3j development toolkit is composed of:
 - neow3j devpack
 - neow3j compiler
 
-Visit http://neow3j.io for more information on neow3j and the technical documentation.
+Visit [neow3j.io](https://neow3j.io) for more information on neow3j and the technical documentation.
 
 ## Using the *neow3j SDK*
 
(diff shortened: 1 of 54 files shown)


gsmachado (approved):
Good job! It enhanced the code, yes.

I particularly liked `ObjectInitializationTest.java` and `InstanceVariablesTest.java`. I strongly believe that good tests would be very very key in order to extend the compiler in the near future. 😄  💯 

Something to think about: these tests are part of the `test-integration`, which do **not** run with `./gradlew test`. Should we make part of it, or, maybe, run the integration tests in the CI-CD? If yes, let's create an issue for it

gsmachado:
@csmuller Just check my approval comment, and I will merge his PR now.

csmuller:
> Good job! It enhanced the code, yes.
> 
> I particularly liked `ObjectInitializationTest.java` and `InstanceVariablesTest.java`. I strongly believe that good tests would be very very key in order to extend the compiler in the near future. 😄 💯
> 
> Something to think about: these tests are part of the `test-integration`, which do **not** run with `./gradlew test`. Should we make part of it, or, maybe, run the integration tests in the CI-CD? If yes, let's create an issue for it

Because the integration tests have to start up a docker container for each test class it makes the CI-CD run slower. Of course it is safer if we would automatically run them on pull-request. If we can run them only for pull-requests and not for every push I think it makes sense to activate them for CI-CD. I'll create an issue on this.

PR #234, opened 2020-10-15
Generating debugging information

csmuller:


csmuller:
In this current state the `Compiler` has a method that takes a set of class files and a set of source files which are then mapped by simply checking if the fully qualified name of a class is contained in the a source file's path.

gsmachado (approved):
I asked all the questions directly to @csmuller and was all clarified. Ready to merge! 🎉 💯

PR #236, opened 2020-10-19
Adaption of byte array initialization

csmuller:
The initialization of byte arrays should not produce a call to NEWBUFFER and a bunch of SETITEM commands. It has been replaced with PUSHDATA and CONVERTER opcodes.

Also refactored the compiler tests. The testet contracts are now inner classe. This makes the project view a bit more tidy. I.e., there are no functional changes in the changed test classes.


gsmachado (approved):
Good PR! I pretty much liked the tests which were added. 👍 💯

gsmachado:
Merging now.

PR #237, opened 2020-10-21
Improvements, fixes, unit tests for the compiler

csmuller:


compiler/src/main/java/io/neow3j/compiler/AsmHelper.java
```diff
@@ -27,18 +27,42 @@
 
 public class AsmHelper {
 
+    /**
+     * Gets the {@code MethodNode} corresponding to the method called in the given instruction.
+     *
+     * @param methodInsn The method instruction.
+     * @param owner      The class that contains the searched for method.
```
gsmachado, line 34:
```suggestion
     * @param owner      The class that contains the method to be searched.
```

gsmachado (approved):
Just a minor change to be made. All in all, great PR, with good tests. 👍 💯 🎉

PR #240, opened 2020-10-26
Improvements, fixes, unit tests for the compiler

csmuller:
Includes the fix for the string literal conversion that produced wrong offsets in the contract manifest.


compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
```diff
@@ -241,8 +235,8 @@ private static void handleStaticFieldConverter(MethodNode methodNode,
             }
         }
         byte[] newOperand = Arrays.copyOfRange(newInsnBytes, 1, newInsnBytes.length);
-        lastNeoInsn.setOpcode(OpCode.get(newInsnBytes[0]));
-        lastNeoInsn.setOperand(newOperand);
+        callingNeoMethod.replaceLastInstruction(
```
gsmachado, line 238: 💯 👍 🥇

gsmachado (approved):
AWESOME work looking for a needle in a haystack. 👍 💯 🥇 

😄

PR #241, opened 2020-10-27
Support String.length() and correct dynamic byte[] initialization

csmuller:


gsmachado (approved):
Small and up-to-the-point tests. Nice. 💯  👍 
No comments from my side.

Issue #185, opened 2020-07-28
Support multi-file smart contracts (debugging)


csmuller:
The compiler should support smart contracts that are spread over multiple `.java` files.

Because Java does not have functionality like C# provides with the keyword `partial`, another solution is needed for enabling multi-file contracts.  
Our proposal is to use an annotation which has a name or ID attribute. That annotation is then applied to every class that is part of the smart contract and the compiler can determine using the name/ID of that annotation if two or more files belong to the same contract.

gsmachado:
As discussed in #242, we need to discuss (and do) the following:

There are two things to be discussed, still:

- [ ] When the method `.compileClasses(canonicalClassNames, setOfSourceSetFilesPath)` is used, meaning that we give several Smart Contract classes to be compiled, as well as with multiple sources, shouldn't we return also multiple `CompilationUnit` objects?
- [ ] If the answer from the previous question is "no, just one `CompilationUnit` is required", then, how should we infer the NEF, Debug Info, and Manifest file names?

We should have multiple ways to specify Smart Contract class names to be compiled:

For multiple files:

```
neow3jCompiler {
    classNames "io.neow3j.examples.NameService", "io.neow3j.examples.NEP5"
    debug false
}
```

For single file:

```
neow3jCompiler {
    classNames "io.neow3j.examples.NameService"
    debug false
}
```

For multiple files with array:

```
neow3jCompiler {
    classNames = [ "io.neow3j.examples.NameService", "io.neow3j.examples.NEP5" ]
    debug false
}
```

For single file with array:

```
neow3jCompiler {
    classNames = [ "io.neow3j.examples.NameService" ]
    debug false
}
```

For single file, keeping backwards compatibility:

```
neow3jCompiler {
    className = "io.neow3j.examples.NameService"
    debug = false
}
```

csmuller:
I challenge the option I mentioned in this issue's description. I don't anymore consider it a good option to have a contract spread it's public methods over multiple class files. That does not seem intuitive at all. There should only be one main contract class file that holds all public (invokable) contract methods.  

That doesn't mean that all contract functionality has to be in the same file. The compiler should still support multiple files (and classes). The files besides the main contract class contain classes that encapsulate parts of the contract's logic. Custom objects used in the contract for organizing data can also live in their own files.

The step to smart contract libraries support isn't far from here.

gsmachado:
> That does not seem intuitive at all. There should only be one main contract class file that holds all public (invokable) contract methods.
[...]

PR #242, opened 2020-10-28
Feature/gradle debug info

gsmachado:
There are two things to be discussed, still:

- [ ] When the method `.compileClasses(canonicalClassNames, setOfSourceSetFilesPath)` is used, meaning that we give several Smart Contract classes to be compiled, as well as with multiple sources, shouldn't we return also multiple `CompilationUnit` objects?
- [ ] If the answer from the previous question is "no, just one `CompilationUnit` is required", then, how should we infer the NEF, Debug Info, and Manifest file names?

Now, there are multiple ways to specify Smart Contract class names to be compiled! 🎉 

For multiple files:

```
neow3jCompiler {
    classNames "io.neow3j.examples.NameService", "io.neow3j.examples.NEP5"
    debug false
}
```

For single file:

```
neow3jCompiler {
    classNames "io.neow3j.examples.NameService"
    debug false
}
```

For multiple files with array:

```
neow3jCompiler {
    classNames = [ "io.neow3j.examples.NameService", "io.neow3j.examples.NEP5" ]
    debug false
}
```

For single file with array:

```
neow3jCompiler {
    classNames = [ "io.neow3j.examples.NameService" ]
    debug false
}
```

For single file, keeping backwards compatibility:

```
neow3jCompiler {
    className = "io.neow3j.examples.NameService"
    debug = false
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -153,10 +153,12 @@ subprojects {
 
     artifacts { archives sourcesJar, javadocsJar, testsJar }
 
-    // Avoid to publish gradle-plugin module.
-    // See "gradle-plugin/README.md" for Gradle Plugin release procedures
-    if (!project.name.startsWith('gradle-plugin') &&
-            !project.name.startsWith(rootProject.name)) {
+    // We just avoid to publish the root project.
+    // The gradle-plugin module is fine and desirable to publish to JFrog/SonaType.
+    // In addition the gradle-plugin module should be published to the Gradle Plugin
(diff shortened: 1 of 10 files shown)


gsmachado:
@csmuller even though the PR is in draft mode, I would like that you check what was done, and take a look at my comment.

csmuller (approved):

gsmachado:
Let's keep the branch for support us on #185 👍 🎉

PR #251, opened 2020-11-05
Fixes a bug with variables, parameters or return types of a custom type

csmuller:
It was not possible to compile a contract having a variable, parameter or return type of a custom class.


gsmachado (approved):
I liked the tests in the `ObjectsTest.java` file. 😃 
Not much to comment on this one.

PR #258, opened 2020-11-10
Devpack helper methods

csmuller:
Cover helper methods with unit tests.
Added a couple of new methods that manipulate strings and already existed for byte arrays.


gsmachado (approved):
@csmuller great work! 🎉 
short PR, small tests representing functionalities 👌

Issue #249, opened 2020-11-02
Fixes, improvements, refactorings for the compiler and devpack


csmuller:
- [ ] Boolean expressions are converted to `JMP<eq>` + `PUSH<n>` statements by the neow3j compiler. But using `PUSH` opcodes creates integers instead of boolean `StackItem`s. The neon compiler on the other hand uses, e.g., the `NUMEQUAL` opcode which puts a `Boolean` to the stack. The neow3j compiler needs to behave the same. Otherwise it can happen that the returned types are not matching the information given in the manifest file. The same problem is present in the neon compiler (see issue neo-project/neo#1912).
- [ ] When creating a new array, e.g., `new boolean[10]`, the neow3j compiler uses the `NEWARRAY` neo-vm opcode. As a result the items in the array have the initial stack item type `ANY` which doesn't necessarily lead to errors but is definitely not optimal. Make the compiler use `NEWARRAY_T` instead when the type of the array is clear.

csmuller:
Created two specific issues from this and added them to the backlog. See #272 and #273.

Issue #181, opened 2020-07-28
Support exceptions


csmuller:
The compiler should support Java exceptions to the extend that their functionality is mappable to NeoVM code.
Keywords: `throw`, `try`, `catch`, `finally`.

csmuller:
Functionality and tests implemented in PR #268.

PR #259, opened 2020-11-11
compiler support for exception throwing

csmuller:
This is only about throwing exceptions not about the try-catch-finally blocks.

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
+++ b/compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
@@ -9,6 +9,7 @@ import static io.neow3j.compiler.Compiler.addInstruction;
 import static io.neow3j.compiler.Compiler.addPushNumber;
 import static io.neow3j.compiler.Compiler.addReverseArguments;
 import static io.neow3j.compiler.Compiler.addSyscall;
+import static io.neow3j.compiler.Compiler.buildPushDataInsn;
 import static io.neow3j.compiler.Compiler.findSuperCallToObjectCtor;
 import static io.neow3j.compiler.Compiler.handleInsn;
 import static io.neow3j.compiler.LocalVariableHelper.buildStoreOrLoadVariableInsn;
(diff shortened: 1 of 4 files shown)


gsmachado:
> This is only about throwing exceptions not about the try-catch-finally blocks.

Clear. 👍 

Short question: how complex is the try-catch-finally blocks structure translation to NeoVM?

compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
```diff
@@ -247,9 +315,9 @@ private static AbstractInsnNode convertConstructorCall(TypeInsnNode typeInsn,
     }
 
     // Checks if the given instruction is a call to the given classes constructor (i.e., <init>).
-    private static boolean isCallToCtor(AbstractInsnNode insn, ClassNode owner) {
+    private static boolean isCallToCtor(AbstractInsnNode insn, String ownerInternalName) {
```
gsmachado, line 318: This change was necessary just to make the comparison in an easier way, right?

Or are there more reasons? Just trying to understand.

csmuller:
> Short question: how complex is the try-catch-finally blocks structure translation to NeoVM?

That's not a short question ;)
I don't have the full picture yet. Will have it till the end of the day. Then I can tell you more. But it seems possible in the foreseen time.

csmuller:
This pull-request can be merged independent of the try-catch functionality.

gsmachado (approved):
Good stuff! 👍 

So, just for my understanding:

By reading the `CompilerExceptionsTest` and `ThrowExceptionTest` class, I could get the limitations of the exception/throw mechanism.
Am I wrong, or it means that only `Exception` is allowed to be instantiated and thrown?

compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
csmuller: Yes, just to make the method usable in more places. Only the name of the owner is needed inside of the method anyway, so why pass the whole ClassNode.

csmuller:
> Am I wrong, or it means that only `Exception` is allowed to be instantiated and thrown?

Indeed, only `Exception` is supported. We can re-evaluate this restriction after implementing the try-catch functionality. Maybe we'll allow the usage of other, e.g. custom exceptions, as well.

gsmachado:
@csmuller great.

It would be nice if we keep these "nice-to-have" features somewhere. Either as an issue or in your ToDos. haha.

Then we could propose further extensions and/or enhancements in the near future. 👍 💯

csmuller:
I have it on my todo list for the moment. We can transfer it to #249 if we find that it makes sense when finishing with #181.

gsmachado:
> I have it on my todo list for the moment. We can transfer it to #249 if we find that it makes sense when finishing with #181.

That's a good approach, thanks. 👍

PR #268, opened 2020-11-24
Support for exceptions and try-catch-finally blocks

csmuller:


compiler/src/main/java/io/neow3j/compiler/DebugInfo.java
```diff
@@ -117,8 +118,12 @@ public static DebugInfo buildDebugInfo(CompilationUnit compUnit) {
     private static List<String> collectVars(Collection<NeoVariable> vars) {
         List<String> varStrings = new ArrayList<>();
         for (NeoVariable var : vars) {
+            String name = var.getName();
+            if (name == null) {
```
csmuller, line 122: This is under construction. There can be variables that appear in ASM but are actually not variables in the Java source code. Therefore, they don't have a name and should not appear in the debug info.

gsmachado (approved):
Cool! I loved the tests. 💯  ❤️

PR #271, opened 2020-11-30
Feature/contract events

csmuller:


devpack/src/main/java/io/neow3j/devpack/events/Event10Args.java
```diff
@@ -0,0 +1,7 @@
+package io.neow3j.devpack.events;
+
+public class Event10Args<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10> implements Event {
+
+    public native void send(T1 arg1, T2 arg2, T3 arg3, T4 arg4, T5 arg5, T6 arg6, T7 arg7, T8 arg8,
```
csmuller, line 5: @gsmachado What do you think about this method name?
In the code one would use events like this
```
transferEvent.send("transfer happened");
```
Should it be something else than send, like trigger?

gsmachado: @csmuller in my opinion, [trigger](https://www.thefreedictionary.com/trigger) is more appropriate for events then [send](https://www.thefreedictionary.com/send).

Unless you point me to some other examples (from other repos) that use `.send()`, explicitly.

gsmachado (approved):
Good stuff! 💯 

As usual, the best part are: `ContractEventsIntegrationTest` and `ContractEventsTest` classes. 😃 🥇 🎉

devpack/src/main/java/io/neow3j/devpack/events/Event10Args.java
gsmachado: I will wait until you rename this method, push, then I resolve this, and I merge it. 👍

csmuller: Actually the method name `notify()` would make sense because the syscall is called `notify` anyway. "Notify" is also used, e.g., in neo-go.

gsmachado: yes, I agree with `.notify()`. Even better.

gsmachado (approved):
💪 💯 🎉 👍

PR #294, opened 2020-12-07
Fix for a bug in the event compilation

csmuller:
The compilation of events with primitive array type parameters, e.g., `Event2Args<byte[], Integer>` had a bug. The type parameters weren't parsed correctly leading to errors in the manifest and debugging information.


gsmachado (approved):
As usual, I get in love with test cases. 😄  ❤️ 
Not that lastIndexOf was not also seen, haha, but the tests always immediately get my eyes. 💯

PR #297, opened 2020-12-08
Improve and test devpack's Enumerator and Iterator

csmuller:


gsmachado (approved):
I will not get tired to say how much I value classes like `EnumeratorTest`, `CompilerTest`, `IteratorTest`, etc... 💯  🍾  🥇 

Merging. 👍

PR #301, opened 2020-12-10
Cover the Blockchain class with tests

csmuller:


gsmachado (approved):
Tests = Great stuff! 😄

PR #355, opened 2021-01-18
Setting of contract name in contract manifest

csmuller:


compiler/src/main/java/io/neow3j/compiler/CompilationUnit.java
```diff
@@ -42,8 +38,8 @@
      */
     private DebugInfo debugInfo;
 
-    // The set of classes that represent the compiled contract.
```
csmuller, line 45: Since we are not properly supporting multi-file contracts at the moment, I switched this back to a single contract class variable. This is necessary to know from which class the compiler should extract the name.
Multiple changes in this PR are due to this change here.

gsmachado: @csmuller agreed. 💯 

We have an issue open for that, so, it's fine.

gsmachado (approved):
It's great! Let's move on! 👍

PR #358, opened 2021-01-19
Fix filenames of NEF and debug info files from gradle plugin output

csmuller:
The filenames of the NEF and debug info file are not set according to the name attribute in the manifest. 
I refactored the whole file writing process a bit.


gsmachado (approved):
Great stuff! 💪  👍

PR #361, opened 2021-01-21
Added JavaDoc to RPC call methods

mialbu:
Further, renamed `witnesses` to `signers` in `invokeFunction` method and adapted the mapping lambda expression.


gsmachado (approved):
good work! 💯

PR #362, opened 2021-01-21
TxBuilder invalid script handling and fixing minor multi-sig signing bug

### LINKED ISSUE #350: Throw exception if transaction script fails/faults
For example, in `TransactionBuilder` class, method `getSystemFeeForScript()`, we don't check if the `response` has a `FAULT` state. If it has, throw an exception and handle it up in the stack.

mialbu:
Resolves #350 #364


gsmachado (approved):
Cool! Good fix. 💪

contract/src/test/resources/responses/invokescript_exception.json
```diff
@@ -0,0 +1,11 @@
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "script": "DA5PcmFjbGVDb250cmFjdEEa93tn",
+    "state": "FAULT",
+    "gasconsumed": "0.0099915",
```
gsmachado, line 7: Can you check other .json files for the tests? I think that now the `gasconsumed` attribute doesn't return decimal anymore. Please, adapt.

Also, make sure that in your tests you check if the correct amount of fees/gas added. This helps us, since if something changes, then this test fails. 😄

core/src/main/java/io/neow3j/protocol/core/methods/response/InvocationResult.java
```diff
@@ -49,6 +49,10 @@ public String getState() {
         return state;
     }
 
+    public boolean hasStateFault() {
+        return state.equals("FAULT");
```
gsmachado, line 53: Is there any possibility that the `state` is `null`? Does it make sense to add this check here?

gsmachado (changes requested):
Some changes to be made.

contract/src/test/resources/responses/invokescript_exception.json
mialbu: I updated several tests now with the newest responses.

core/src/main/java/io/neow3j/protocol/core/methods/response/InvocationResult.java
mialbu: I don't think so. The state is always either `NONE`, `HALT`, `FAULT` or `BREAK`.

gsmachado (approved):
Now it's good to go! 🚀

PR #366, opened 2021-01-22
Candidate check in NeoToken vote method

### LINKED ISSUE #351: Add check if candidate voted for with NeoToken.vote() exists
Throw an exception if the user is voting for a public key that is not a candidate (checked with `NeoToken.getCandidates()`).

mialbu:
Closes #351

### DIFF
--- a/contract/src/main/java/io/neow3j/contract/NeoToken.java
+++ b/contract/src/main/java/io/neow3j/contract/NeoToken.java
@@ -228,32 +228,44 @@ public class NeoToken extends Nep17Token {
     }
 
     /**
-     * Creates a transaction script to vote for the given validators and initializes a {@link
-     * TransactionBuilder} based on this script.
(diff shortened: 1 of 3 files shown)


gsmachado (approved):

gsmachado:
Before I merge: is there an integration test that could cover this check? Or do you think it's not necessary?

mialbu:
I've tested manually with the neo3-privatenet-docker. Since we don't have integration tests in the contract module, there are no integration tests for this. However, we could setup integration tests for the contract module in general.

gsmachado:
> I've tested manually with the neo3-privatenet-docker. Since we don't have integration tests in the contract module, there are no integration tests for this. However, we could setup integration tests for the contract module in general.

Ok. Let's discuss setting up integration tests for the contract module in the near future, yes. 💪 

I'm going to merge now. 👍

PR #372, opened 2021-01-27
Add @Safe annotation

### LINKED ISSUE #314: Remove 'savemethods' from ContractManifest and add a '@Safe' annotation as a replacement
The `safemethods` attribute was removed from the contract manifest and moved into the method declaration in the ABI.
- Adapt the `ContractManifest` class and all affected code locations to reflect this change. 
- Implement a new devpack annotation that can be used on smart contract methods to signal that they are safe. 
- Adapt the compiler to support this annotation, such that it adds the information to the ABI.

csmuller:
Closes #314.

### DIFF
--- a/compiler/src/test/java/io/neow3j/compiler/SafeMethodTest.java
+++ b/compiler/src/test/java/io/neow3j/compiler/SafeMethodTest.java
@@ -1,0 +1,39 @@
+package io.neow3j.compiler;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import io.neow3j.devpack.annotations.Safe;
+import io.neow3j.protocol.core.methods.response.ContractManifest.ContractABI.ContractMethod;
+import java.io.IOException;
+import java.util.List;
+import org.junit.Test;
+
+public class SafeMethodTest {
+
+    @Test
+    public void methodShouldBeMarkedSafeIfAnnotatedWithSafeAnnotation() throws IOException {
+        CompilationUnit unit = new Compiler().compileClass(SafeMethodTestContract.class.getName());
+        List<ContractMethod> methods = unit.getManifest().getAbi().getMethods();
+        assertTrue(methods.get(0).isSafe());
+    }
+
+    @Test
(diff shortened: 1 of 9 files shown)


compiler/src/test/java/io/neow3j/compiler/SafeMethodTest.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Test
+    public void methodShouldNotBeMarkedSafeIfNotAnnotatedWithSafeAnnotation() throws IOException {
+        CompilationUnit unit = new Compiler().compileClass(SafeMethodTestContract.class.getName());
+        List<ContractMethod> methods = unit.getManifest().getAbi().getMethods();
+        assertFalse(methods.get(1).isSafe());
+    }
+
+    static class SafeMethodTestContract {
```
gsmachado, line 28: @csmuller just for the sake of completeness: can you add a test case where a contract has at least two `@Safe` methods? 😄 

Moreover, is there any restriction for this `@Safe` annotation? Like, for example, all other annotations can be added if `@Safe` is present, right?

gsmachado (changes requested):

compiler/src/test/java/io/neow3j/compiler/SafeMethodTest.java
csmuller: > just for the sake of completeness: can you add a test case where a contract has at least two @Safe methods?

Will do.

> Moreover, is there any restriction for this @Safe annotation? Like, for example, all other annotations can be added if @Safe is present, right?

Good thinking. I found one point. It doesn't make sense to use the annotation on methods that are not public, i.e. that  don't show up in the manifest. I will throw an exception if the annotation is used on non-public methods.

I don't see conflicts with other annotations like `OnDeployment` or `OnVerification`. Just because methods marked with such annotations are special methods it doesn't mean that the developer cannot mark them as "safe".

gsmachado (approved):
Now it's great! 😃  💪 
I will merge it. 👍

PR #392, opened 2021-01-29
Add Groups, Trusts, and Permissions to the contract manifest

### LINKED ISSUE #348: Add support for trusts and permissions
A contract manifest contains the attributes `trusts`, and `permissions`. The compiler must offer a way to specify these attributes when implementing a Java smart contract.

### LINKED ISSUE #354: Support adding contract groups to contract manifest
The contract manifest has a `group` attribute which is a list of `Group` objects that represent a trusted circle of contracts which can be interacted with without any extra notifications. I.e. contracts in the group can be called by the contract and can call the contract without triggering a notification. 

The neow3j devpack needs to allow setting of such group objects in the contract manifest

gsmachado:
Closes #348 
Closes #354

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
+++ b/compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
@@ -2,13 +2,17 @@ package io.neow3j.compiler;
 
 import static io.neow3j.compiler.AsmHelper.getAnnotationNode;
 import static io.neow3j.compiler.AsmHelper.hasAnnotations;
+import static java.util.Optional.ofNullable;
 
 import io.neow3j.contract.ContractParameter;
 import io.neow3j.devpack.annotations.DisplayName;
+import io.neow3j.devpack.annotations.Group.Groups;
 import io.neow3j.devpack.annotations.ManifestExtra;
 import io.neow3j.devpack.annotations.ManifestExtra.ManifestExtras;
+import io.neow3j.devpack.annotations.Permission.Permissions;
 import io.neow3j.devpack.annotations.Safe;
 import io.neow3j.devpack.annotations.SupportedStandards;
+import io.neow3j.devpack.annotations.Trust.Trusts;
 import io.neow3j.model.types.ContractParameterType;
 import io.neow3j.protocol.core.methods.response.ContractManifest;
 import io.neow3j.protocol.core.methods.response.ContractManifest.ContractABI;
@@ -18,7 +22,7 @@ import io.neow3j.protocol.core.methods.response.ContractManifest.ContractGroup;
 import io.neow3j.protocol.core.methods.response.ContractManifest.ContractPermission;
 import io.neow3j.utils.ClassUtils;
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -31,6 +35,7 @@ import org.objectweb.asm.tree.ClassNode;
 /**
  * Contains all functionality required to build a contract manifest from a compilation unit.
  */
+@SuppressWarnings("unchecked")
 public class ManifestBuilder {
 
     public static ContractManifest buildManifest(CompilationUnit compUnit) {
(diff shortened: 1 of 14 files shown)


compiler/src/test/java/io/neow3j/compiler/GroupManifestTest.java
```diff
@@ -0,0 +1,76 @@
...
+        assertThat(groups.get(1).getSignature(), is("signature2"));
+    }
+
+    @Test
+    public void withoutGroupsAnnotation() throws IOException {
+        CompilationUnit unit = new Compiler().compileClass(GroupManifestTestContractWithoutAnnotation.class.getName());
+        List<ContractGroup> groups = unit.getManifest().getGroups();
+        assertThat(groups, hasSize(0));
+    }
+
```
csmuller, line 48: Please add a test case that uses only one `@Group` annotation.

compiler/src/test/java/io/neow3j/compiler/PermissionManifestTest.java
```diff
@@ -0,0 +1,84 @@
...
+    public void withoutPermissionsAnnotation() throws IOException {
+        CompilationUnit unit = new Compiler()
+                .compileClass(PermissionManifestTestContractWithoutAnnotation.class.getName());
+        List<ContractPermission> permissions = unit.getManifest().getPermissions();
+        assertThat(permissions, hasSize(1));
+        assertThat(permissions.get(0).getContract(), is("*"));
+        assertThat(permissions.get(0).getMethods(), hasSize(1));
+        assertThat(permissions.get(0).getMethods().get(0), is("*"));
+    }
+
```
csmuller, line 55: Please add a test case that uses only one `@Permission` annotation.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
```diff
@@ -106,17 +110,92 @@ private static ContractABI buildABI(NeoModule neoModule) {
...
+                methods.addAll((List<String>) methodsValues);
+            }
+
+            ContractPermission contractPermission = new ContractPermission(contract, methods);
+            permissions.add(contractPermission);
+        }
+        return permissions;
+    }
+
+    private static List<ContractGroup> getContractGroups(AnnotationNode ann) {
```
csmuller, line 178: We could enforce rules on the annotation values here.
- The public key has to be a 33 bytes long hex string.
- The signature has to be Base64 encoded.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
```diff
@@ -106,17 +110,92 @@ private static ContractABI buildABI(NeoModule neoModule) {
...
+            groups.add(cgManifest);
         }
-        AnnotationNode ann = annotationNode.get();
-        List<String> standards = new ArrayList<>();
-        for (Object standard : (List<?>) ann.values.get(1)) {
-            standards.add((String) standard);
+        return groups;
+    }
+
+    private static List<String> getContractTrusts(AnnotationNode ann) {
```
csmuller, line 191: We could enforce rules on the annotation values here.
The value has to be either a 33 (public key) or 20 bytes long (script hash) hex string.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
```diff
@@ -106,17 +110,92 @@ private static ContractABI buildABI(NeoModule neoModule) {
...
+                .map(ann -> {
+                    List<String> values = new ArrayList<>();
+                    for (Object value : (List<?>) ann.values.get(1)) {
+                        values.add((String) value);
+                    }
+                    return values;
+                });
+    }
+
+    private static List<ContractPermission> getContractPermissions(AnnotationNode ann) {
```
csmuller, line 153: We could enforce rules on the annotation values here.
The contract value has to be either a 33 (public key) or 20 bytes long (script hash) hex string.

csmuller (changes requested):

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
gsmachado: Good point. I will check that. 👍

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
gsmachado: Yes, good point. I will do that.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
gsmachado: Yet another one. 😄 
I will implement it.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
gsmachado: implemented.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
gsmachado: implemented.

compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
gsmachado: implemented.

csmuller (approved):

PR #398, opened 2021-02-04
NefFile updates

csmuller:


csmuller:
The compiler integration tests are still failing because the contract deployment doesn't work yet. We'll fix that in another PR.

gsmachado (approved):
All fine and clear. It even increased a bit the test coverage, which is always good. 💪  👍

PR #399, opened 2021-02-04
Revert gas notation in RPC methods to integers instead of decimals

csmuller:


gsmachado (approved):
Hopefully, this is the last time we change that. 🤯  💪

Issue #270, opened 2020-11-26
Support Map ContractParameter


csmuller:
The neow3j SDK does not yet have support for the Map ContractParameter. Add support for it in `ContractParameter` and `ScriptBuilder` so it can be used for contract calls.

csmuller:
Solved in #468.

PR #400, opened 2021-02-05
Adapt contract calls

csmuller:


utils/src/main/java/io/neow3j/contract/ScriptBuilder.java
```diff
@@ -132,6 +137,7 @@ public ScriptBuilder pushParam(ContractParameter param) {
             case ARRAY:
                 pushArray((ContractParameter[]) value);
                 break;
+            // TODO: Add a case for type MAP.
```
gsmachado, line 140: do we have an issue with the Map?

csmuller: This is tracked in #270.

gsmachado (approved):
💪  👍  💯  Let's move on.

PR #403, opened 2021-02-05
Update management contract

csmuller:


csmuller:
Everything is implemented but the tests for the `ContractManagement` methods `update` and `destroy` in the devpack are not working. I have no idea what the problem is at the moment. This will require some tedious debugging.

gsmachado (approved):
Nothing supercritical to be changed. We're good to go!

As usual, test increasing! I love it. ohohhoho 😄

PR #406, opened 2021-02-08
Verification method integration tests

### LINKED ISSUE #376: invokeContractVerify requires open wallet containing signers
The `invokecontractverify` RPC method requires that the signers added to the call are available in the open wallet on the neo-node.
- [x] Adapt Javadoc of the `Neow3j.invokeContractVerify(...)` method.
- [x] Adapt integration tests in `VerificationMethodIntegrationTest` which should be failing because the wallet is not opened explicitely.

mialbu:
Resolves #376.


gsmachado (approved):
It's good to go! 🚀  👍  👌  💪

Issue #412, opened 2021-02-10
Update LedgerContract


mialbu:
The method `getBlock` now includes 9 instead of 8 entries in the returned array stack item. There may be additional changes as well.

csmuller:
Solved in #468.

Issue #390, opened 2021-01-29
Add the LedgerContract to the devpack


csmuller:
In preview5, Neo has moved a lot of block and transaction related functionality to the new `LedgerContract`. The neow3j devpack needs to provide access to this contract for contract developers.

- [ ] Add the `LedgerContract` as a contract interface
- [ ] Adapt other parts of the devpack which have been changed because of the introduction of the ledger contract.

gsmachado:
@csmuller is this done with PR #417 , right?

I'll go ahead and close the issue. 💪  If something's still missing, please re-open.

PR #416, opened 2021-02-11
Fix compiler and compiler tests

csmuller:


csmuller:
The rest of the integration tests will be fixed when introducing the LedgerContract to the devpack.

gsmachado (approved):
Cool! Ready to go from my side. 🚀

gsmachado:
> The rest of the integration tests will be fixed when introducing the LedgerContract to the devpack.

But those test fixes (the remaining ones) should be fixed part of the other issues (#390 #412), right?

csmuller:
> But those test fixes (the remaining ones) should be fixed part of the other issues (#390 #412), right?

Yes.

PR #418, opened 2021-02-15
Oracle contract for devpack

csmuller:


csmuller:
- [ ] In order for the integration tests of the OracleContract to work the Oracle Service needs to run on the privatnet.
To run the Oracle service, a candidate needs to be registered (`NeoToken.registerCandidate`) and designated under the Oracle Role (`RolenManagement.designateAsRole`). Then, the oracle service can be started on the node with `start oracle`.

gsmachado:
> * [ ]  In order for the integration tests of the OracleContract to work the Oracle Service needs to run on the privatnet.
>   To run the Oracle service, a candidate needs to be registered (`NeoToken.registerCandidate`) and designated under the Oracle Role (`RolenManagement.designateAsRole`). Then, the oracle service can be started on the node with `start oracle`.

Cool. Interesting.

Let's keep in mind that this case could be a nice example to add to the `neo3-examples` git repo. 😃

compiler/src/test-integration/java/io/neow3j/compiler/OracleContractTest.java
```diff
@@ -71,32 +69,39 @@ public void performRequest() throws Throwable {
         Await.waitUntilTransactionIsExecuted(response.getSendRawTransaction().getHash(), neow3j);
 
         // Invoke contract that requests data from oracle.
-        String url = "http://127.0.0.1:98239/test"; // Request URL that is not reachable
+        String url = "http://127.0.0.1:64928/test"; // Request URL that is not reachable
```
gsmachado, line 72: sometimes this port can be available (open), and sometimes not? Or due to the wire mock it should be always unavailable?

If it's always unavailable, it's fine to leave like this. However, if it's not enforced by wire mock, then I suggest to add a random port therefor  every time the test runs.

csmuller: I added WireMock to actually mock the web service called by the oracle.
But I didn't manage to make it work fully. The request of the oracle to the web service times out.
It doesn't matter for this test, because we don't need to test the whole Oracle process just the `OracleContract`.

compiler/src/test-integration/java/io/neow3j/compiler/OracleContractTest.java
```diff
@@ -1,67 +1,129 @@
...
+                .sign().send();
+        Await.waitUntilTransactionIsExecuted(response.getSendRawTransaction().getHash(), neow3j);
+        // Designate as oracle.
+        response = new RoleManagement(neow3j).designateAsRole(Role.ORACLE, Arrays.asList(publicKey))
+                .wallet(wallet)
+                .signers(Signer.calledByEntry(committee.getScriptHash()))
+                .sign().send();
+        Await.waitUntilTransactionIsExecuted(response.getSendRawTransaction().getHash(), neow3j);
+
+        // Start the oracle service on the neo-node
```
gsmachado, line 81:
```suggestion
        // TODO: Remove this when https://github.com/neo-project/neo-modules/issues/523
        // gets implemented.
        // Start the oracle service on the neo-node
```

contract/src/test/java/io/neow3j/contract/GasTokenTest.java
```diff
@@ -14,19 +14,13 @@
...
 
     private final static String GASTOKEN_SCRIPTHASH = "70e2301955bf1e74cbb31d18c2f96972abadb328";
 
     @Before
     public void setUp() {
-        // Configuring WireMock to use default host and the dynamic port set in WireMockRule.
-        int port = this.wireMockRule.port();
-        WireMock.configureFor(port);
-        neow = Neow3j.build(new HttpService("http://127.0.0.1:" + port));
+        neow = Neow3j.build(new HttpService("http://127.0.0.1:8080"));
```
gsmachado, line 23: the port is not supposed to be dynamic?

gsmachado (changes requested):

contract/src/test/java/io/neow3j/contract/GasTokenTest.java
csmuller: The service is not actually used by all of the tests in that class. I should make that more clear.

gsmachado (approved):
Well-done figuring all of this out. 😄

PR #421, opened 2021-02-15
Devpack integer type normalization and package refactor

csmuller:
Additional to the standardized use of `int` (instead of long or short) throughout the devpack I also removed the packages `neo` and `system` which did not make much sense in my opinion. The dotnet devpack still has this separation, but I don't see a meaningful separation in it from the view of the developer.


csmuller:
The integration tests fail because of the OracleContract, which is fixed in #418.

devpack/src/main/java/io/neow3j/devpack/Storage.java
```diff
@@ -80,16 +80,6 @@
...
 
-    /**
-     * Stores the given key-value pair.
-     *
-     * @param context The storage context to store the value in.
-     * @param key     The key.
-     * @param value   The value to store.
-     */
-    @Syscall(SYSTEM_STORAGE_PUT)
-    public static native void put(StorageContext context, byte[] key, long value);
```
gsmachado, line 91: @csmuller the only thing about this pull request that I didn't get, is: why did you decide to remove these methods in the `Storage.java` class?

gsmachado (approved):
Good refactoring and clean-up. 💪  👍 

I just posted one question that I didn't get. Then we can merge it,

devpack/src/main/java/io/neow3j/devpack/Storage.java
csmuller: They are duplicates of the same methods with the only difference of using `int` instead of `long` as the parameters. Since the goal of this issue/PR was to switch everything to `int`, I decided to remove the `long`-methods.

PR #422, opened 2021-02-15
Convenience Method for Array ContractParameter

### LINKED ISSUE #296: Add `ContractParameter` methods for conveniently creating array parameters
Currently the SDK allows to create `Array` contract parameters by passing other `ContractParameter` instances into the `ContractParameter.array(...)` method. That is not very convenient when the developer needs to pass a long list of integers or other objects.

Add methods for passing arrays and lists of objects that create a valid `Array` contract parameter.
E.g., `ContractParameter.array(new int[]{1, 2, 3})` should be possible.

mialbu:
Closes #296


gsmachado (changes requested):
@mialbu could you also add test cases for

- `array()`
- `array(null)`

After this, I can approve the PR. 😄

gsmachado (approved):
Cool! Good one, increased test coverage! 💪  👍  🚀

PR #425, opened 2021-02-16
Get contract state by name

### LINKED ISSUE #375: getContractState by name of native contracts
`getcontractstate` now allows fetching the contract state of native contracts by name (additionally to contract hash).
Neow3j's `getContractState` method already takes a string as an argument.

I think we should add an overload method that takes a `ScriptHash` and use the already existing method to cover the new use case, i.e., simply change the method documentation and the parameter name.

mialbu:
Closes #375


gsmachado (approved):
Cool! I think this implementation is quite nice and flexible. 💪  👍  💯

PR #426, opened 2021-02-17
Use magic number from `getversion` for NeoConfig

csmuller:
I moved the magic number and the address version to the Neow3j class, because it is at the center of all blockchain interactions and can easily be used as a replacement for `NeoConfig`. Therefore, `NeoConfig` became obsolete.


gsmachado (approved):
Great! Good work. This simplifies things quite a bit. 👍  😄

PR #430, opened 2021-02-22
Changes in transaction signing to allow "signing" with a smart contract

csmuller:


core/src/main/java/io/neow3j/transaction/Transaction.java
```diff
@@ -169,18 +169,14 @@ public String getTxId() throws IOException {
...
-
-        for (Signer signer : this.getSigners()) {
-            if (!witnesses.contains(signer.getScriptHash())) {
-                throw new TransactionConfigurationException("The transaction does not have a "
-                        + "signature for each of its signers.");
-            }
+        if (getSigners().size() != getWitnesses().size()) {
+            throw new TransactionConfigurationException("The transaction does not have the same " +
+                    "number of signers and witnesses. For every signer there has to be one " +
+                    "witness, even if that witness is empty.");
```
csmuller, line 179: Important change to review.

core/src/main/java/io/neow3j/transaction/Transaction.java
```diff
@@ -298,7 +294,6 @@ private void serializeWithoutWitnesses(BinaryWriter writer) throws IOException {
     @Override
     public void serialize(BinaryWriter writer) throws IOException {
         serializeWithoutWitnesses(writer);
-        this.witnesses.sort(Comparator.comparing(Witness::getScriptHash));
```
csmuller, line 301: Sorting obviously doesn't work if there are empty witnesses. I also didn't find the sorting anymore in the current neo code.

csmuller, line 301: Sorting obviously doesn't work if there are empty witnesses. I also didn't find the sorting anymore in the current neo code.

core/src/main/java/io/neow3j/transaction/Witness.java
```diff
@@ -23,7 +24,13 @@
...
     private ScriptHash scriptHash;
 
+    /**
+     * Constructs an empty witness with a zero-valued script hash.
+     */
     public Witness() {
+        invocationScript = new InvocationScript();
+        verificationScript = new VerificationScript();
+        scriptHash = ScriptHash.ZERO;
     }
```
csmuller, line 34: Making sure that an empty witness is instantiated with empty byte arrays and a zero-valued script hash

contract/src/main/java/io/neow3j/contract/TransactionBuilder.java
```diff
@@ -391,56 +385,42 @@ private long getSystemFeeForScript() throws IOException {
...
-        Transaction tx = new Transaction(neow, version, nonce, validUntilBlock, signers, 0,
-                0, attributes, script, new ArrayList<>());
-        getSignerAccounts().forEach(s -> {
-            tx.addWitness(new Witness(new byte[]{}, s.getVerificationScript().getScript()));
-        });
-        NeoCalculateNetworkFee networkFeeResult = neow
-                .calculateNetworkFee(Numeric.toHexStringNoPrefix(tx.toArray())).send();
-        BigInteger networkFee = networkFeeResult.getNetworkFee().getNetworkFee();
-        return networkFee.longValue();
-    }
```
csmuller, line 424: Important change to review.

contract/src/main/java/io/neow3j/contract/TransactionBuilder.java
```diff
@@ -391,56 +385,42 @@ private long getSystemFeeForScript() throws IOException {
...
-        });
-        return sigAccounts;
+        }
+        if (!hasAtLeastOneSigningAccount) {
+            throw new TransactionConfigurationException("No signers were set for which an account" +
+                    " with verification script exists in the wallet.");
+        }
+        String txHex = Numeric.toHexStringNoPrefix(tx.toArray());
+        return neow.calculateNetworkFee(txHex).send().getNetworkFee().getNetworkFee().longValue();
     }
```
csmuller, line 424: Important change to review.

csmuller:
Sorry for the many formatting changes. The autoformatter hit hard. I marked the code locations that are important.

contract/src/main/java/io/neow3j/contract/TransactionBuilder.java
gsmachado: Alright. I think this is the logic I was expecting! Cool. 💪

core/src/main/java/io/neow3j/transaction/Transaction.java
gsmachado: Yes. Good. This change is done since now we need to add "empty" witnesses.

One use case we could test in the future:
Is there a use case that two empty witnesses are added (like... two signers without priv keys)? If so, the Neo node would accept that? Or both would feed to be translated into a SINGLE empty witness?

This is not required for now, but it would be good to know....

core/src/main/java/io/neow3j/transaction/Transaction.java
gsmachado: very interesting...... so, if order doesn't matter, my comment above (about merging two empty witnesses) would make more or less sense? 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔 🤔

core/src/main/java/io/neow3j/transaction/Witness.java
gsmachado: Good!

gsmachado (approved):

core/src/main/java/io/neow3j/transaction/Transaction.java
csmuller: If a transaction transfers tokens from two different smart contracts that would lead to exactly the scenario you described. The two contracts would be added as signers and two empty witnesses have to be added (and probably a signature from the contract owner, depending on the contracts' `verify` methods).

core/src/main/java/io/neow3j/transaction/Transaction.java
csmuller: Good point. The empty witnesses are anyway undistinguishable and could therefore be merged into one. I was not enlightened about this by the Neo code yet. I would need to read a bit more.

core/src/main/java/io/neow3j/transaction/Transaction.java
gsmachado: @csmuller should we create a GitHub issue for this one?
Also, about investigating whether combining multiple empty Witnesses into one is the right way of doing things or not.

If you open an issue about that, then we could merge this PR.

csmuller: Alright, I created issue #431 for this problem.

Issue #311, opened 2020-12-11
Static constructor with local variable leads to exception


csmuller:
The neow3j compiler allows the usage of static constructors, but a `NullPointerException` occurs when using local variables in the static constructor. An example contract is shown in the code example below.

```
public class ExampleContract {

    private static int integer;

    static {
        int i = 10;
        integer = i * 2;
    }

    public static int main() throws Exception {
        return integer;
    }
}
```
We assume that the problem only arises because the neo-vm doesn't allow local variables in the INITSSLOT method. That opcode does not allow the definition of local variables.
The solution is to either
(1) build a parser that parses the static constructor at compile time and resolves values of static class variables already at compile time, or 
(2) we simply disallow the usage of local variables in static constructors and throw a compiler exception if the developer uses them.

gsmachado:
@csmuller I think we can close this one, right? 👍

PR #435, opened 2021-02-25
Prohibit local variables in static constructor

csmuller:
The solution to #311 is that we prohibit usage of local variables in static constructors. The neo-vm does not support them (in the INITSSLOT method, which is the corresponding construct of a static constructor.

The alternative solution would be to parse the static constructor and try to produce neo-vm code without using local variables. Creating such a parser obviously implies much higher effort than the solution of this PR.

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/Compiler.java
+++ b/compiler/src/main/java/io/neow3j/compiler/Compiler.java
@@ -58,7 +58,6 @@ public class Compiler {
 
     public static final String INSTANCE_CTOR = "<init>";
     private static final String CLASS_CTOR = "<clinit>";
-    private static final String INITSSLOT_METHOD_NAME = "_initialize";
     public static final String THIS_KEYWORD = "this";
 
     public static final String INSTRUCTION_ANNOTATION_OPERAND = "operand";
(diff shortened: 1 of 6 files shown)


gsmachado:
> The alternative solution would be to parse the static constructor and try to produce neo-vm code without using local variables. Creating such a parser obviously implies much higher effort than the solution of this PR.

Yes. Agree.

I think there's no real and great use case for this yet. That's why I'm in favor of the implementation you've come up with in this PR.

gsmachado (approved):
I think that's the way to go, yes. Also, as far as I understood, we even got a new "feature" (support super constructors) with this PR. Or is that more a fix?

Anyway, I think it's acceptable to go in this direction. Good stuff. 💪  👍

csmuller:
> I think that's the way to go, yes. Also, as far as I understood, we even got a new "feature" (support super constructors) with this PR. Or is that more a fix?

It's like an easing of the restrictions. The developer is allowed to extend from other classes. But that doesn't mean that we support inheritance fully yet. E.g., if the super constructor that is called in a sub-classes constructer has custom code in it, that code is not processed by the compiler.

PR #437, opened 2021-02-25
Name Service wrapper class

### LINKED ISSUE #419: Add the NameService to the SDK
In preview5, a name service was added to the native contracts as a non-fungible token.
- [x] Add a wrapper of the native NameService contract to the SDK.

For detailed information about the methods of the native NameService contract, see [here](https://docs.neo.org/v3/docs/en-us/reference/scapi/fw/dotnet/neo/NameService.html).

mialbu:
Closes #419

### DIFF
--- a/contract/src/main/java/io/neow3j/contract/NeoNameService.java
+++ b/contract/src/main/java/io/neow3j/contract/NeoNameService.java
@@ -1,5 +1,12 @@
 package io.neow3j.contract;
 
+import static io.neow3j.contract.ContractParameter.byteArray;
+import static io.neow3j.contract.ContractParameter.hash160;
+import static io.neow3j.contract.ContractParameter.integer;
+import static io.neow3j.contract.ContractParameter.string;
+import static io.neow3j.model.types.StackItemType.MAP;
+import static java.util.Collections.singletonList;
+
 import io.neow3j.contract.exceptions.UnexpectedReturnTypeException;
 import io.neow3j.protocol.Neow3j;
 import io.neow3j.protocol.core.RecordType;
@@ -9,19 +16,11 @@ import io.neow3j.protocol.core.methods.response.NameState;
 import io.neow3j.protocol.core.methods.response.StackItem;
 import io.neow3j.utils.Numeric;
 import io.neow3j.wallet.Wallet;
-import org.bouncycastle.util.IPAddress;
 
 import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
-
-import static io.neow3j.contract.ContractParameter.byteArray;
-import static io.neow3j.contract.ContractParameter.hash160;
-import static io.neow3j.contract.ContractParameter.integer;
-import static io.neow3j.contract.ContractParameter.string;
-import static io.neow3j.model.types.StackItemType.MAP;
-import static java.util.Collections.singletonList;
 
 /**
  * Represents the NameService native contract and provides methods to invoke its functions.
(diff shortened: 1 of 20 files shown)


contract/src/main/java/io/neow3j/contract/NeoNameService.java
```diff
@@ -0,0 +1,368 @@
...
+        if (!shouldBeAvailable && isAvailable) {
+            throw new IllegalArgumentException("The domain name '" + name + "' is not registered.");
+        }
+    }
+
+    // checks that the domain name matches the required regex and contains two levels.
+    private void checkDomainNameValidity(String name) {
+        checkRegexMatch(NAME_REGEX_PATTERN, name);
+        if (name.split("\\.").length != 2) {
+            throw new IllegalArgumentException("Only second domain names are allowed to be " +
```
gsmachado, line 170:
```suggestion
            throw new IllegalArgumentException("Only second-level domain names are allowed to be " +
```

contract/src/main/java/io/neow3j/contract/NeoNameService.java
```diff
@@ -0,0 +1,368 @@
...
+        }
+    }
+
+    /**
+     * Creates a transaction script to update the TTL of the domain name and initializes a
+     * {@link TransactionBuilder} based on this script.
+     * <p>
+     * Each call will extend the validity period of the domain name by one year.
+     * <p>
+     * Only supports to renew the second domain name.
```
gsmachado, line 189:
```suggestion
     * Only supports renewing the second-level domain name.
```

contract/src/main/java/io/neow3j/contract/NeoNameService.java
```diff
@@ -0,0 +1,368 @@
...
+     *
+     * @return the price to register a domain.
+     * @throws IOException if there was a problem fetching information from the Neo node.
+     */
+    public BigInteger getPrice() throws IOException {
+        return callFuncReturningInt(GET_PRICE);
+    }
+
+    /**
+     * Checks if the specified second domain name is available.
```
gsmachado, line 125:
```suggestion
     * Checks if the specified second-level domain name is available.
```

contract/src/main/java/io/neow3j/contract/NeoNameService.java
```diff
@@ -0,0 +1,368 @@
...
+    private static final String SET_RECORD = "setRecord";
+    private static final String GET_RECORD = "getRecord";
+    private static final String DELETE_RECORD = "deleteRecord";
+    private static final String RESOLVE = "resolve";
+
+    private static final String PROPERTIES = "properties";
+
+    private static final BigInteger MAXIMAL_PRICE = new BigInteger("1000000000000");
+    private static final Pattern ROOT_REGEX_PATTERN = Pattern.compile("^[a-z][a-z0-9]{0,15}$");
+    public static final Pattern NAME_REGEX_PATTERN = Pattern.compile(
```
gsmachado, line 50: Any reason why this attribute is `public`? If no, leave it private, or change all `Pattern` attributes to protected (if required on tests).

contract/src/main/java/io/neow3j/contract/NeoNameService.java
```diff
@@ -0,0 +1,368 @@
...
+     * @param owner the address of the domain owner.
+     * @return a transaction builder.
+     */
+    public TransactionBuilder register(String name, ScriptHash owner) throws IOException {
+        checkDomainNameAvailability(name, true);
+        return invokeFunction(REGISTER, string(name), hash160(owner));
+    }
+
+    // checks if a domain name is available
+    private void checkDomainNameAvailability(String name, boolean shouldBeAvailable)
```
gsmachado, line 155: Maybe it'd be a good idea to change `shouldBeAvailable` to `isForRegistration`.

I think `isForRegistration` it's a bit more legible than `shouldBeAvailable` since `true` or `false` depends if it's for `renew` (or related actions) or for `register`.

gsmachado (changes requested):
Minor changes only! 😄 

Great work. 💪  🚀

gsmachado:
@mialbu once those are solved, I will happily merge. Good set of tests as well. 👍

contract/src/main/java/io/neow3j/contract/NeoNameService.java
mialbu: No, I missed changing it back to private. I'll change it.

contract/src/main/java/io/neow3j/contract/NeoNameService.java
mialbu: I agree, I'll change it to `isForRegistration`.

gsmachado (approved):
Great! 💪

gsmachado:
@mialbu last question: do you think it would make sense to have very basic integration tests for `NameService`?

Like that we would be able to detect if they change something in the signature of the contract methods, and/or functionality. What do you think? We don't need to have an extensive integration test coverage, but just maybe `register`, `isAvailable`, `getPrice`...

Let us know before I merge.

mialbu:
> @mialbu last question: do you think it would make sense to have very basic integration tests for `NameService`?

Yes, I thought about integration tests as well, however, imo we should generally setup integration tests for the contract module.
I think this should be a separate task to tackle. I could open issues for this. I would open one for read and one for write integration tests, as we have it in the core module. What do you think?

gsmachado:
> I think this should be a separate task to tackle. I would suggest creating two issues for this, one for read and one for write integration tests as we have it in the core module. What do you think?

Ok, @mialbu, let's proceed like this: please, create the issues and I will move on and merge this task now. Assign the issues to you, and put them on `ToDo`, please. Leave it without Milestone for the moment.

mialbu:
@gsmachado I'll open one single issue, maybe it makes sense to create integration test files per wrapper class in the contract module. We can still discuss this and split it to multiple separate issues when the time has come to tackle these integration tests.

PR #438, opened 2021-03-01
Smart Contract Array Initialization

### LINKED ISSUE #273: The compiler should initialize arrays with a type
When creating a new array, e.g., new boolean[10], the neow3j compiler uses the NEWARRAY neo-vm opcode. As a result the items in the array have the initial stack item type ANY which doesn't necessarily lead to errors but is definitely not optimal. Make the compiler use NEWARRAY_T with a type operand in cases where the type of the array is clear.

- Parse additional information on JVM's `NEWARRAY` and `ANEWARRAY` instructions to figure out the type of the array
- Initialize empty arrays with the neo-vm opcode `NEWARRAY0` because it is much cheaper.

csmuller:
Closes #273.


compiler/src/main/java/io/neow3j/compiler/converters/ArraysConverter.java
```diff
@@ -41,7 +50,8 @@ public AbstractInsnNode convert(AbstractInsnNode insn, NeoMethod neoMethod,
                 if (isByteArrayInstantiation(insn)) {
                     insn = handleNewByteArray(insn, neoMethod);
                 } else {
-                    neoMethod.addInstruction(new NeoInstruction(OpCode.NEWARRAY));
+                    handleNewArray(insn, neoMethod);
+                    // TODO!!!
```
gsmachado, line 54: What do you mean by this `TODO` here? Can you explain a bit more in detail?

compiler/src/test-integration/java/io/neow3j/compiler/ArraysConverterIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+            return new String[5];
+        }
+
+        public static String getEmptyStringFromCreatedArray() {
+            String[] s = new String[5];
+            return s[3];
+        }
+
+        public static byte[][] createArrayOfByteArrays() {
+            return new byte[][]{new byte[]{0, 1, 3}, new byte[]{0}};
```
gsmachado, line 112: cool!

gsmachado (approved):
Cool! Good work. Good tests. 💪 👍

compiler/src/main/java/io/neow3j/compiler/converters/ArraysConverter.java
csmuller: Removed it.

PR #444, opened 2021-03-05
Added Hash256 class and adapted general usage of Hash160 and Hash256

### LINKED ISSUE #254: Rename ScriptHash to Hash160 and add Hash256 to align neow3j with other Neo projects
Other Neo SDKs and devpacks use the `UInt160` type to represent script hashes, i.e., hashes produced with RIPEMD, and `UInt256` for hashes produced from SHA-256. Neow3j uses the `ScriptHash` class for `UInt160` and does not have any type for `UInt256`.

We should adapt neow3j to use `UInt160` instead of `ScriptHash` and add `UInt256`.

**Update after discussion** (see below):
The class `ScriptHash` will be renamed to `Hash160` and a new class `Hash256` will be added.
- [x] Rename `ScriptHash` to `Hash160`
- [x] Add `Hash256`

mialbu:
Closes #254

### DIFF
--- a/compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
@@ -1,6 +1,7 @@
 package io.neow3j.compiler;
 
 import static io.neow3j.contract.ContractParameter.string;
+import static java.util.Collections.singletonList;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
@@ -10,7 +11,7 @@ import io.neow3j.devpack.annotations.OnVerification;
 import io.neow3j.devpack.Runtime;
 import io.neow3j.protocol.core.methods.response.NeoInvokeContractVerify;
 import io.neow3j.transaction.Signer;
-import java.util.Arrays;
+
 import org.junit.BeforeClass;
 import org.junit.Test;
 
@@ -21,14 +22,25 @@ public class VerificationMethodIntegrationTest extends ContractTest {
         setUp(VerificationMethodIntegrationTestContract.class.getName());
         // The RPC method invokecontractverify requires an open wallet on the
         // neo-node.
-        neow3j.openWallet("wallet.json","neo").send();
+        neow3j.openWallet("wallet.json", "neo").send();
     }
 
     @Test
     public void callVerifyWithContractOwner() throws Throwable {
         NeoInvokeContractVerify response = neow3j
                 .invokeContractVerify(contract.getScriptHash().toString(),
-                        Arrays.asList(string("hello, world!")),
+                        singletonList(string("hello, world!")),
+                        Signer.calledByEntry(defaultAccount.getScriptHash()))
+                .send();
+
+        assertTrue(response.getInvocationResult().getStack().get(0).asBoolean().getValue());
+    }
+
+    @Test
+    public void callVerifyWithContractOwner_fromString() throws Throwable {
+        NeoInvokeContractVerify response = neow3j
+                .invokeContractVerify(contract.getScriptHash(),
+                        singletonList(string("hello, world!")),
                         Signer.calledByEntry(defaultAccount.getScriptHash()))
                 .send();
 
@@ -39,7 +51,18 @@ public class VerificationMethodIntegrationTest extends ContractTest {
     public void callVerifyWithOtherSigner() throws Throwable {
         NeoInvokeContractVerify response = neow3j
                 .invokeContractVerify(contract.getScriptHash().toString(),
-                        Arrays.asList(string("hello, world!")),
+                        singletonList(string("hello, world!")),
+                        Signer.calledByEntry(committee.getScriptHash()))
+                .send();
+
+        assertFalse(response.getInvocationResult().getStack().get(0).asBoolean().getValue());
+    }
+
+    @Test
+    public void callVerifyWithOtherSigner_fromString() throws Throwable {
+        NeoInvokeContractVerify response = neow3j
+                .invokeContractVerify(contract.getScriptHash().toString(),
+                        singletonList(string("hello, world!")),
                         Signer.calledByEntry(committee.getScriptHash()))
                 .send();
 
@@ -58,4 +81,5 @@ public class VerificationMethodIntegrationTest extends ContractTest {
         }
 
     }
+
 }
--- a/compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
@@ -50,7 +50,7 @@ public class VerificationMethodIntegrationTest extends ContractTest {
     @Test
     public void callVerifyWithOtherSigner() throws Throwable {
         NeoInvokeContractVerify response = neow3j
-                .invokeContractVerify(contract.getScriptHash().toString(),
+                .invokeContractVerify(contract.getScriptHash(),
                         singletonList(string("hello, world!")),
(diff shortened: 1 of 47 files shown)


core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
```diff
@@ -941,9 +1094,22 @@ public JsonRpc2_0Neow3j(
      */
     @Override
     public Request<?, NeoGetApplicationLog> getApplicationLog(String txId) {
+        return getApplicationLog(new Hash256(txId));
+    }
+
+    /**
+     * Gets the contract event information based on the specified transaction
+     * id. The contract event information is stored under the ApplicationLogs
+     * directory.
```
csmuller, line 1103:
```suggestion
     * Gets the application logs of the specified transaction.
```

Apply the same change to the javadoc of the method above.

core/src/test/java/io/neow3j/contract/Hash160Test.java
```diff
@@ -32,18 +33,8 @@ public void createFromValidHash() {
...
-    }
-
-    @Test(expected = IllegalArgumentException.class)
-    public void createFromTooLongHash() {
-        new Hash160("c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9ba");
-    }
-
-    @Test(expected = IllegalArgumentException.class)
-    public void createFromMidSizedHash() {
-        new Hash160("23ba2703c53263e8d6e522dc32203339dcd8eee9938a3e");
```
csmuller, line 46: Why were these tests removed?

csmuller (changes requested):

core/src/test/java/io/neow3j/contract/Hash160Test.java
gsmachado: I also believe they should remain in the test suite.

compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
```diff
@@ -21,14 +22,25 @@ public static void setUp() throws Throwable {
...
         // neo-node.
-        neow3j.openWallet("wallet.json","neo").send();
+        neow3j.openWallet("wallet.json", "neo").send();
     }
 
     @Test
     public void callVerifyWithContractOwner() throws Throwable {
         NeoInvokeContractVerify response = neow3j
                 .invokeContractVerify(contract.getScriptHash().toString(),
-                        Arrays.asList(string("hello, world!")),
```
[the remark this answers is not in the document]
mialbu: With this pr, the api is extended, such that many methods now also accept `Hash160` or `Hash256` instead of just `String`.
One test covers `invokeContractVerify(String, ...)` and one covers `invokeContractVerify(Hash160, ...)`. Ultimately, the one taking a String will call the other one, but I think it still makes sense to have one test for all methods separately.

core/src/main/java/io/neow3j/contract/ContractParameter.java
```diff
@@ -248,6 +248,16 @@ public static ContractParameter hash256(String hashHexString) {
         return hash256(Numeric.hexStringToByteArray(hashHexString));
     }
 
+    /**
+     * Creates a hash256 parameter from the given hash.
+     *
+     * @param hash a 256-bit hash in little-endian order.
```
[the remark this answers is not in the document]
mialbu: I copied that from the method below, I'll remove it for both.

core/src/test/java/io/neow3j/contract/Hash160Test.java
mialbu: The first two tests don't throw the expected exception because the length is incorrect, but because the hex regex pattern does not match them. Given the test names, I think that is not what is supposed to be tested here. But you're correct, I forgot to replace them accordingly. I'll add more precise tests for that.

For the third test: I don't see the purpose of the `..fromMidSizedHash`. It actually just tests the same as the test with a hash that is too long. Can you tell me, what it was supposed to test, since the naming confused me here?

csmuller: > For the third test: I don't see the purpose of the ..fromMidSizedHash. It actually just tests the same as the test with a hash that is too long. Can you tell me, what it was supposed to test, since the naming confused me here

You're right. You can remove that case. It doesn't test anything else then the `...TooLongHash()` test.

core/src/main/java/io/neow3j/contract/ContractParameter.java
```diff
@@ -248,10 +248,20 @@ public static ContractParameter hash256(String hashHexString) {
...
+     * @return the contract parameter.
+     */
+    public static ContractParameter hash256(Hash256 hash) {
+        return hash256(hash.toArray());
+    }
+
     /**
      * Creates a hash256 parameter from the given bytes.
      *
-     * @param hash bytes (possibly a 256-bit hash) in little-endian order.
```
csmuller, line 254: Here the mention of the endianness should not be removed because the parameter is a plain byte array.

core/src/main/java/io/neow3j/contract/Hash160.java
```diff
@@ -153,16 +155,15 @@ public static Hash160 fromScript(String script) {
 
     private void checkAndThrowHashLength(byte[] scriptHash) {
```
csmuller, line 156: This method still uses the `scriptHash` name in variable and exception message but should use just `hash`.

compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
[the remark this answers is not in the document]
csmuller: 👍 
In that cas the test `callVerifyWithContractOwner()` should not do 
```
.invokeContractVerify(contract.getScriptHash().toString(),
```
but
```
.invokeContractVerify(contract.getScriptHash(),
```
right?

csmuller (changes requested):

core/src/main/java/io/neow3j/contract/ContractParameter.java
mialbu: Yes, you're right, I'll change it back.

gsmachado (approved):

core/src/main/java/io/neow3j/contract/Hash160.java
```diff
@@ -151,18 +153,17 @@ public static Hash160 fromScript(String script) {
...
     }
 
-    private void checkAndThrowHashLength(byte[] scriptHash) {
-        if (scriptHash.length != NeoConstants.HASH160_SIZE) {
-            throw new IllegalArgumentException("Script hash must be " +
-                NeoConstants.HASH160_SIZE + " bytes long but was " + scriptHash.length +
-                " bytes.");
+    private void checkAndThrowHashLength(byte[] hash) {
+        if (hash.length != NeoConstants.HASH160_SIZE) {
+            throw new IllegalArgumentException("Script hash must be " + NeoConstants.HASH160_SIZE +
```
csmuller, line 158:
```suggestion
            throw new IllegalArgumentException("Hash must be " + NeoConstants.HASH160_SIZE +
```

compiler/src/test-integration/java/io/neow3j/compiler/VerificationMethodIntegrationTest.java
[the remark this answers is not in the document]
csmuller: The two tests `callVerifyWithOtherSigner()` and `callVerifyWithOtherSigner_fromString()` still contain the exact same code.

csmuller (changes requested):

csmuller (approved):

PR #445, opened 2021-03-08
Refactoring of stack item API

### LINKED ISSUE #337: Add additional abstractions on `StackItem` and the stack returned by `InvocationResult`
Instead of making the developer cast a stack item to a specific type we should allow retrieving the value immediately as a specific Java type. E.g., instead of requiring `stackItem.asBoolean().getValue()`, return a boolean value immediately for `stackItem.asBoolean()`. Basically, this demands that we remove the layer in which we cast the generic stack item to a specific stack item.
The specific stack item classes, like `ArrayStackItem` or `ByteArrayStackItem` become obsolete and are replaced with direct getter methods on the `StackItem` class.

Furthermore, we need a shortcut for getting stack items from the stack returned by the `InvocationResult`. Consider something like the following two:
`response.getInvocationResult().getStack().getFirst()`
`response.getInvocationResult().getFirstItemFromStack()`

csmuller:
Closes #337


gsmachado (approved):
Awesome work! 💪  👍

Issue #449, opened 2021-03-08
Update IPv4 and IPv6 regex in the NameService wrapper class


mialbu:
See according commit in the Neo core [here](https://github.com/neo-project/neo/commit/e2c68778c921c7ba6d49a1b7bcce627b4f6d3228).

mialbu:
This was implemented in pr #459.

PR #459, opened 2021-03-09
Non-fungible token changes

### LINKED ISSUE #447: Remove description from NameState of the NameService
See according commit in the Neo core [here](https://github.com/neo-project/neo/commit/0d4632821f3f049e065c3b97324b165d9eab9927).

mialbu:
Closes #447 #449

### DIFF
--- a/contract/src/test/java/io/neow3j/contract/NeoNameServiceTest.java
+++ b/contract/src/test/java/io/neow3j/contract/NeoNameServiceTest.java
@@ -379,11 +379,11 @@ public class NeoNameServiceTest {
         setUpWireMockForInvokeFunction(IS_AVAILABLE, "invokefunction_returnFalse.json");
 
         byte[] expectedScript = new ScriptBuilder().contractCall(NeoNameService.SCRIPT_HASH, SET_RECORD,
-                asList(string("client1.neo"), integer(28), string("1234:0:0:0:0:0:0:1234")))
+                asList(string("client1.neo"), integer(28), string("1234::1234")))
                 .toArray();
 
         TransactionBuilder b = new NeoNameService(neow)
-                .setRecord("client1.neo", RecordType.AAAA, "1234:0:0:0:0:0:0:1234");
+                .setRecord("client1.neo", RecordType.AAAA, "1234::1234");
         assertThat(b.getScript(), is(expectedScript));
     }
 
@@ -392,7 +392,7 @@ public class NeoNameServiceTest {
         setUpWireMockForInvokeFunction(IS_AVAILABLE, "invokefunction_returnFalse.json");
         expectedException.expect(IllegalArgumentException.class);
         expectedException.expectMessage("input does not match the required regex.");
-        new NeoNameService(neow).setRecord("client1.neo", RecordType.AAAA, "1234::1234");
+        new NeoNameService(neow).setRecord("client1.neo", RecordType.AAAA, "12345::2");
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


contract/src/test/java/io/neow3j/contract/NeoNameServiceTest.java
```diff
@@ -380,11 +379,11 @@ public void setRecord_typeAAAA() throws IOException {
         setUpWireMockForInvokeFunction(IS_AVAILABLE, "invokefunction_returnFalse.json");
 
         byte[] expectedScript = new ScriptBuilder().contractCall(NeoNameService.SCRIPT_HASH, SET_RECORD,
-                asList(string("client1.neo"), integer(28), string("1234:0:0:0:0:0:0:1234")))
+                asList(string("client1.neo"), integer(28), string("1234::1234")))
```
gsmachado, line 382: if we don't have a test for it, yet, I would create a test case for `1234:0:0:0:0:0:0:1234` as well. To make sure.

core/src/main/java/io/neow3j/protocol/core/methods/response/NameState.java
```diff
@@ -4,20 +4,20 @@
 
 import java.util.Objects;
 
-public class NameState extends TokenState {
+public class NameState extends NFTokenState {
 
     @JsonProperty(value = "expiration")
-    private Integer expiration;
+    private long expiration;
```
gsmachado, line 10: I think it's better to have complex types (in this case, `Long`), since the object has a `@JsonProperty`, and the JSON attribute "expiration" can be `null` for some weird reason. The goal is to not break the whole code because of a problematic serialization/deserialization. Keep this in mind for the future.

gsmachado (changes requested):
Just some changes, more related to tests. 😅

contract/src/test/java/io/neow3j/contract/NeoNameServiceTest.java
mialbu: Actually, I didn't need to change it in the first place, since the regex is hardcoded here and in the neo-core. I just liked to change it because the `::` didn't work on the previous regex in the neo-core. ^^
So I can change it back or add another test, but it won't add better coverage or anything.

core/src/main/java/io/neow3j/protocol/core/methods/response/NameState.java
mialbu: You're right, I'll adapt that.

contract/src/test/java/io/neow3j/contract/NeoNameServiceTest.java
gsmachado: Regex tend to be different across some languages... I don't know if dotnet's version (or regex standard) is the same as JVM-based projects. That's why I asked.

I think that adding an additional test wouldn't hurt.

core/src/main/java/io/neow3j/protocol/core/methods/response/NameState.java
```diff
@@ -4,20 +4,20 @@
...
     @JsonProperty(value = "expiration")
-    private Integer expiration;
+    private Long expiration;
 
     public NameState() {
     }
 
-    public NameState(String name, String description, Integer expiration) {
-        super(name, description);
+    public NameState(String name, long expiration) {
```
gsmachado, line 15:
```suggestion
    public NameState(String name, Long expiration) {
```

gsmachado (approved):
💪  💯

PR #460, opened 2021-03-09
Several RC1 changes for which we don't have a specific issue

csmuller:


csmuller:
The integration test don't work yet because they don't use up to date test containers. We should anyway merge it in order to have these changes available in all the other branches we are opening for RC1 updates.

compiler/src/test-integration/java/io/neow3j/compiler/LedgerContractIntegrationTest.java
```diff
@@ -156,13 +156,11 @@ public void currentHash() throws IOException {
...
 
-    @Ignore("Waiting for the implementation of the LedgerContract in the contract module.")
     @Test
     public void getHash() throws IOException {
         NeoInvokeFunction response = callInvokeFunction();
-        // TODO: Get the script hash of the LedgerContract form the contract module.
-//      assertThat(response.getInvocationResult().getStack().get(0).getHexString(),
-//              is(io.neow3j.contract.LedgerContract.SCRIPT_HASH.toString()));
+        assertThat(response.getInvocationResult().getStack().get(0).getHexString(),
+                is("da65b600f7124ce6c79950c1772a36403104f2be"));
```
gsmachado, line 163: did you use the hex string here on purpose? I think it's better if we start to use something like `LedgerContract.SCRIPT_HASH.toString()`... what do you think?

This applies to other tests.

contract/src/test/java/io/neow3j/contract/NefFileTest.java
```diff
@@ -54,8 +54,10 @@
             "contracts/TestContractWithMethodTokens.nef";
     private static final String TESTCONTRACT_WITH_TOKENS_SCRIPT = "213701004021370000405700017840";
     private final static List<MethodToken> TESTCONTRACT_METHOD_TOKENS = asList(
-            new MethodToken(NeoToken.SCRIPT_HASH, "getGasPerBlock", 0, true, CallFlags.ALL),
-            new MethodToken(GasToken.SCRIPT_HASH, "totalSupply", 0, true, CallFlags.ALL));
+            new MethodToken(new Hash160("f61eebf573ea36593fd43aa150c055ad7906ab83"),
```
gsmachado, line 57: same comment from above applies here, for example.

devpack/src/main/java/io/neow3j/devpack/Storage.java
```diff
@@ -160,224 +152,4 @@
     public static native Iterator<Entry<String, byte[]>> find(StorageContext context,
             String prefix);
 
-    /**
```
gsmachado, line 163: can you briefly explain to me what exactly happened with all the Storage context and all the other calls? Will they be moved? or removed, completely?

gsmachado (changes requested):
(minor) Clarifications and potential changes. Once those are settled, we can merge. 💪

compiler/src/test-integration/java/io/neow3j/compiler/LedgerContractIntegrationTest.java
csmuller: I use `SomeContract.SCRIPT_HASH.toString()` wherever possible. But in this case there is no `LedgerContract` in the `contract` package, therefore, i used the hex string directly.

devpack/src/main/java/io/neow3j/devpack/Storage.java
csmuller: Good point. So, the implicit storage context methods were removed without replacement in the dotnet-devpack, which  I use as a reference implementation. That said, I agree that those methods did not have much use. No developer would use such methods because they always incur the overhead of fetching the `StorageContext` before doing the actual storage operation. That is more expensive than using the explicit storage context methods.

PR #475, opened 2021-03-18
Replace BlockParameter with BigInteger

csmuller:
In the RC1 branch I removed the usage of BlockParameter in the `Neow3j`/`JsonRpc2_0Neow3j` classes because they were a bit cumbersome to use instead of directly using BigIntegers. I now went ahead and remove the `BlockParameter`, `BlockParameterIndex` and `BlockParameterName` altogether. I hope no feelings are hurt with these changes 🐱


gsmachado:
I'll only accept this PR because of your emoji. 🤣 

Jokes aside, I think it's a good move. 👍  I support this idea.

gsmachado (approved):
It's all fine! I agree with the changes. I think it's simpler now, and more clear.

PR #477, opened 2021-03-19
Integration test config changes

mialbu:
Adapted the test container setup to specify the config file to be used. The block time in the read only integration tests is now set to five seconds to make some tests more robust, e.g. `testGetProof()`.


csmuller (approved):
All good. 😼

PR #479, opened 2021-03-22
Add method to subscribe to new blocks without catching up

csmuller:
This PR adds only a convenience method to subscribe to new incoming blocks.


gsmachado (approved):
good one 💪

PR #491, opened 2021-03-31
Adapt interop services

csmuller:


csmuller:
Not mergable because it depends on changes that will be introduced with RC2.

PR #492, opened 2021-03-31
Uniform hash endianness handling

### LINKED ISSUE #442: Make the handling of endianness in hashes (e.g. script hashes) uniform
The Hash160 and Hash256 classes store the hash values as a little endian byte array but when `toString()` is called the value is returned as big-endian. Make this uniform and use only one byte ordering (I lean towards big-endian because that is how Java mostly handles binary values).

Additionally, check for the usage of hashes throughout neow3j and make sure that their endianness is consistent and well documented.

Check the Neo documentation about endianness [here](https://docs.neo.org/v3/docs/en-us/basic/concept/endian.html).

mialbu:
The hashes in the classes `Hash160` and `Hash256` are now stored in big-endian format and its constructors now all take big-endian formats as parameter. This way the hashes are uniformly handled in big-endian format.

Resolves #442.

### DIFF
--- a/core/src/main/java/io/neow3j/contract/Hash160.java
+++ b/core/src/main/java/io/neow3j/contract/Hash160.java
@@ -9,6 +9,7 @@ import static io.neow3j.utils.Numeric.hexStringToByteArray;
 import static io.neow3j.utils.Numeric.isValidHexString;
 import static io.neow3j.utils.Numeric.toHexStringNoPrefix;
 
+import com.fasterxml.jackson.annotation.JsonValue;
 import io.neow3j.constants.NeoConstants;
 import io.neow3j.io.BinaryReader;
 import io.neow3j.io.BinaryWriter;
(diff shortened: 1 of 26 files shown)


core/src/main/java/io/neow3j/contract/Hash160.java
```diff
@@ -129,18 +145,15 @@ public static Hash160 fromAddress(String address) {
      * @return the script hash.
      */
     public static Hash160 fromScript(byte[] script) {
-        // There is no need to reverse the hash. The hashing method returns the script hash in
-        // little-endian format.
-        return new Hash160(Hash.sha256AndThenRipemd160(script));
+        return new Hash160(reverseArray(sha256AndThenRipemd160(script)));
```
csmuller, line 148: I don't trust this anymore. As far as i can tell the `sha256AndThenRipemd160` method is returning big-endian, but in the javadoc of that method i wrote it returns the hash in little-endian ordering. I want to run some tests on this branch.

csmuller (changes requested):

core/src/main/java/io/neow3j/contract/Hash160.java
csmuller: Seems fine, according to my tests.

csmuller (approved):

PR #498, opened 2021-04-06
VM state enum

gsmachado:


core/src/test/java/io/neow3j/protocol/core/ResponseTest.java
```diff
@@ -1637,7 +1638,32 @@ public void testInvokeFunction_withoutOrEmpty_Params() {
...
+
+    @Test
+    public void testInvokeFunction_emptyState() {
+        buildResponse(
+                "{\n" +
+                        "    \"jsonrpc\": \"2.0\",\n" +
+                        "    \"id\": 1,\n" +
+                        "    \"result\": {\n" +
+                        "        \"script\": \"10c00c0962616c616e63654f660c14897720d8cd76f4f00abfa37c0edd889c208fde9b41627d5b52\",\n" +
+                        "        \"state\": \"\",\n" +
```
mialbu, line 1655: Can this case really occur?

mialbu (commented):
All good :+1:
I have only that one question about the empty state. :grin:

core/src/test/java/io/neow3j/protocol/core/ResponseTest.java
gsmachado: Not sure, but I think that's the reason why they have `NONE`. 😄 

I would prefer to leave this test like this, even if it's not realistic. It tests the serialization/deserialization of the JSON to `NONE` state. 👍

PR #507, opened 2021-04-09
multi file compilation

csmuller:


gsmachado:
I'm really looking forward to this one! 💪  👍  💯

csmuller:
This PR introduces the basics for debugging of multi-file Java smart contracts. The code should be extendable to also include  source lookups for libraries used in a smart contract.

compiler/src/main/java/io/neow3j/compiler/sourcelookup/ISourceContainer.java
```diff
@@ -0,0 +1,18 @@
...
+public interface ISourceContainer {
+
+    /**
+     * Searches this container for the source file(s) with the given name. Returns multiple files
+     * if multiple items with the same name are found.
+     *
+     * @param name The name of the file.
+     * @return a list of sources files found matching the given name.
+     */
+    File findSourceFile(String name);
```
gsmachado, line 16: In the javadoc above it mentions multiple files, but here it's just `FIle` and not a `List<File>`. Should we change to be a bit more flexible and implement as a `List`?

devpack/src/main/java/io/neow3j/devpack/Iterator.java
```diff
@@ -15,11 +15,13 @@
...
  * example:
  * <pre>
- * {@literal ...}
- * {@literal List<ByteString> list = new io.neow3j.devpack.List<>()};
- * {@literal iterator.next()};
- * {@literal list.add(iterator.getValue());}
- * {@literal ...}
+ * {@literal
+ * ...
+ *  List<ByteString> list = new io.neow3j.devpack.List<>();
```
gsmachado, line 20: Please, revert to `@literal`. 

The `@literal` is used to avoid warnings about encoding and strange chars like `<` and `>` in the JavaDoc.

If you use `./gradlew javadoc` you will see all the related warnings.

gsmachado (changes requested):
Just some comments and minor changes.

devpack/src/main/java/io/neow3j/devpack/Iterator.java
csmuller: I think this was a merge conflict at some point...
Anyway, the `{@literal}` is wrapped around the whole code example. I checked with `./graldew javadoc` and it builds without issues.

gsmachado: OH, my fault! I haven't seen that the `@literal` was multi-line! Awesome. All fine. 😄

gsmachado (approved):
All fine, and ready to go! 💪  💯

PR #509, opened 2021-04-13
Make transfer amounts uniform

mialbu:


gsmachado:
This PR is great and necessary.

However, it will change quite a bit in our examples and documentation. We need to be pay attention to adapt everything that needs updates (just to keep in mind).

Maybe we need to be proactive and already create an issue to update examples and docs for `3.9.0`?

csmuller (approved):

PR #514, opened 2021-04-15
Changed the the Map.Entry to Iterator.Struct

csmuller:
Since the neo-vm returns a StructStackItem when iterating over map entries, I changed the Map.Entry type to Iterator.Struct.

### DIFF
--- a/compiler/src/test-integration/java/io/neow3j/compiler/IteratorTest.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/IteratorTest.java
@@ -2,7 +2,7 @@ package io.neow3j.compiler;
 
 import io.neow3j.devpack.Iterator;
 import io.neow3j.devpack.Map;
-import io.neow3j.devpack.Map.Entry;
+import io.neow3j.devpack.Iterator.Struct;
 import io.neow3j.protocol.core.methods.response.NeoInvokeFunction;
 import io.neow3j.protocol.core.methods.response.StackItem;
 import org.junit.ClassRule;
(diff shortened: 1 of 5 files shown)


csmuller:
Or should we simply call it Tuple and thereby open up for the possibility to have higher arity tuples in the future?

csmuller:
The integration tests ran successfully in the first run. Three runs were started automatically and i cancelled the second two. That's why it doesn't scho success.

gsmachado:
> The integration tests ran successfully in the first run. Three runs were started automatically and i cancelled the second two. That's why it doesn't scho success.

Strange. I need to investigate why this is happening. Probably is because of the self-hosted GitHub Action runner (a VM within nucbox).

gsmachado:
@csmuller the integration tests are successful if you run locally?

csmuller:
> @csmuller the integration tests are successful if you run locally?

The ran in the github action as well, it just doesn't show because I cancelled the latest run that got started.
I'll manually start it once again so that it shows green.

mialbu:
> > The integration tests ran successfully in the first run. Three runs were started automatically and i cancelled the second two. That's why it doesn't scho success.
> 
> Strange. I need to investigate why this is happening. Probably is because of the self-hosted GitHub Action runner (a VM within nucbox).

The integration test workflow ran three times because the `integration.yml` is triggered on `[assigned, review_requested, ready_for_review]`. When @csmuller opened the pr and requested a review from me and @gsmachado, this workflow was triggered three times: once for his own assignment, once for @gsmachado's review request and once for my review request and since it was never a draft pr, `ready_for_review` did not trigger the workflow.

I think we should change the workflow trigger types for the integration workflow. I'm not sure what combination is best, but e.g. `assigned` is not necessary imo.

Check [this](https://frontside.com/blog/2020-05-26-github-actions-pull_request/) out, it's a really good explanation about when workflows are triggered.

mialbu:
> Or should we simply call it Tuple and thereby open up for the possibility to have higher arity tuples in the future?

Is any storage handled other than a k-v pair in the Neo vm? If so, how? If not, do you think there will be more than k-v pairs in the Neo vm in the near future?

mialbu:
> Or should we simply call it Tuple and thereby open up for the possibility to have higher arity tuples in the future?

I think we should stick to `Struct`. It's more coherent with the Neo vm that way.

mialbu (approved):
:100:

PR #520, opened 2021-04-23
Fix for bug in compilation of try-catch

csmuller:


gsmachado (approved):
Clarified my questions in a bilateral discussion. All fine. I always love the tests! 😄  😍

PR #522, opened 2021-04-23
Handle calls to `Object` methods

csmuller:


devpack/build.gradle
```diff
@@ -2,4 +2,8 @@ description 'neow3j: Java Development Package for Neo Smart Contracts'
 
 dependencies {
     compile project(':contract')
-}
\ No newline at end of file
+}
+
+compileJava {
+    options.debugOptions.debugLevel = "source,lines,vars"
```
gsmachado, line 8: so, this is necessary only for the `devpack` module?

Is this necessary to set in the user's `build.gradle` project as well? (if users wants debug info)

gsmachado (approved):
Great work! 💪  Just asked a minor clarification, but we can take this offline. 💯

PR #524, opened 2021-04-26
Support for instanceof

### LINKED ISSUE #321: Support JVM's `INSTANCEOF` opcode
The JVM opcode `INSTANCEOF` - and Java's `instanceof` that maps to it - nicely map to neo-vm's `ISTYPE` opcode. The neow3j compiler should be extended to support that JVM opcode.

csmuller:
Closes #321.

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/Compiler.java
+++ b/compiler/src/main/java/io/neow3j/compiler/Compiler.java
@@ -119,8 +119,8 @@ public class Compiler {
         if (typeName.equals(Hash256.class.getTypeName())) {
             return ContractParameterType.HASH256;
         }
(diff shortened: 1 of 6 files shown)


gsmachado (changes requested):
Missing points:

- test `instanceof` in a if/else construction
- test `instanceof` in a static variable

csmuller:
> Missing points:
> 
> * test `instanceof` in a if/else construction
> * test `instanceof` in a static variable

I added test cases for those. Let me know if you want more complex ones. Since the INSTANCEOF opcode from the JVM can be mapped one-to-one to the ISTYPE neo-vm opcode I have faith that the compiler will manage complex cases just fine.

gsmachado (approved):
Great stuff! Good to go. 😄  💪  💯

PR #528, opened 2021-04-28
Tests for vararg method parameters

### LINKED ISSUE #275: Support variable-length parameters
The neow3j compiler should support variable-length parameters in smart contract methods. One example where this is required is the `Runtime.notify(Object... objs)` method.  
The Java compiler converts the variable-length parameter into an array and passes that as the argument to the called methods. This already works for primitive types but not for objects (uses the JVM `ANEWARRAY` opcode instead of `NEWARRAY`).

- Implement support for variable-length parameters of non-primitve type.

csmuller:
Closes #275.

### DIFF
--- a/compiler/src/test-integration/java/io/neow3j/compiler/ArraysConverterIntegrationTest.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/ArraysConverterIntegrationTest.java
@@ -1,6 +1,9 @@
 package io.neow3j.compiler;
 
 import io.neow3j.contract.Hash160;
+import io.neow3j.contract.Hash256;
+import io.neow3j.devpack.events.Event1Arg;
+import io.neow3j.protocol.core.methods.response.NeoApplicationLog;
 import io.neow3j.protocol.core.methods.response.NeoInvokeFunction;
 import io.neow3j.protocol.core.methods.response.StackItem;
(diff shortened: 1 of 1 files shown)


compiler/src/test-integration/java/io/neow3j/compiler/ArraysConverterIntegrationTest.java
```diff
@@ -132,6 +167,23 @@ public static String getEmptyStringFromCreatedArray() {
...
+            int sum = integers[0];
+            for (int i : integers) {
+                sum += i;
+            }
+            return sum + a;
+        }
+
+        static Event1Arg<io.neow3j.devpack.Hash160> event;
+
+        public static void callRuntimeNotifyWithVarArgs(io.neow3j.devpack.Hash160... hashes) {
```
gsmachado, line 181: I know it's a bit of an overkill, but can we make a test with an `Object` (complex type) that is not from the Runtime/devpack? (i.e., any complex type, like `String`, or `Integer`)

If this doesn't make sense, then it's fine and we can move on.

gsmachado (changes requested):
Just a minor comment (request for adding a test case) before merging.

gsmachado (approved):
Great! Good to go! 💪 🚀

PR #548, opened 2021-05-16
Empty catch with inner var init

gsmachado:


gsmachado:
Just created some integration tests that are useful.

The compilation of the test class is failing at the moment, since empty catch can't be compiled, somehow, with var initialization in the try block.

gsmachado (commented):
Great! 💪

csmuller (approved):
More test coverage on the compiler 🥳

PR #551, opened 2021-05-19
Fix string switch case bug

csmuller:


gsmachado (approved):
as discussed in our call, great stuff!

💪  💯

PR #554, opened 2021-05-24
Make getHash of Contract Interfaces return little-endian hash

csmuller:
Also, improved the handling of contract interfaces in the compiler.


gsmachado (approved):
Great! In fact, this way is more natural to use (as discussed in Discord).

PR #555, opened 2021-05-24
Make Contract Parameter Constructor protected

csmuller:


gsmachado (approved):
That's actually a good move, yes. 👍  💪  💯

PR #556, opened 2021-05-25
Support divisible NFT methods and added NFT methods returning iterators

mialbu:


contract/src/main/java/io/neow3j/contract/NonFungibleToken.java
```diff
@@ -43,6 +48,43 @@ public NonFungibleToken(Hash160 scriptHash, Neow3j neow) {
...
+                .stream()
+                .map(StackItem::getByteArray)
+                .collect(Collectors.toList());
+    }
+
+    // Non-divisible NFT methods
+
     /**
      * Creates a transaction script to transfer a non-fungible token and initializes a
      * {@link TransactionBuilder} based on this script.
```
csmuller, line 90: Add the following comment here too, as you have added it to the other overload of this method.
"This method is intended to be used for non-divisible NFTs only."

contract/src/main/java/io/neow3j/contract/NonFungibleToken.java
```diff
@@ -92,17 +139,66 @@ public TransactionBuilder transfer(Wallet wallet, Hash160 to, byte[] tokenID,
...
+        if (getDecimals() == 0) {
+            throw new IllegalStateException("This method is only intended for divisible NFTs.");
+        }
+
+        return invokeFunction(TRANSFER, hash160(from), hash160(to), integer(amount),
+                byteArray(tokenID), data)
+                .wallet(wallet)
+                .signers(calledByEntry(from));
+    }
+
```
csmuller, line 178: Please add Javadoc to these methods and consider renaming them to simply `transfer`. I think it's enough if the javadoc says that they are intended for divisible tokens only.

contract/src/main/java/io/neow3j/contract/NonFungibleToken.java
```diff
@@ -92,17 +139,66 @@ public TransactionBuilder transfer(Wallet wallet, Hash160 to, byte[] tokenID,
...
     /**
-     * Gets the balance of the token with {@code tokenID} for the given account.
+     * Gets the owners of the token with {@code tokenId}.
+     * <p>
+     * Consider that for this RPC the returned list may be limited in size and not reveal all
+     * entries that exist on the contract.
+     *
```
csmuller, line 184: Add comment that this is only usable with divisible tokens.

contract/src/main/java/io/neow3j/contract/NonFungibleToken.java
```diff
@@ -117,13 +213,38 @@ public Hash160 ownerOf(byte[] tokenId) throws IOException {
      * @throws UnexpectedReturnTypeException if the contract invocation did not return something
      *                                       interpretable as a number.
      */
-    public BigInteger balanceOf(Hash160 owner)
+    public BigInteger balanceOf(Hash160 owner, byte[] tokenId)
```
csmuller, line 216: Add comment in javadoc that this is only usable with divisible tokens.

contract/src/main/java/io/neow3j/contract/NonFungibleToken.java
```diff
@@ -136,7 +257,7 @@ public NFTokenState properties(byte[] tokenID) throws IOException {
             Map<StackItem, StackItem> map = item.getMap();
 
             return new NFTokenState(
-                    map.get(new ByteStringStackItem("name".getBytes(UTF_8))).getString());
+                    map.get(new ByteStringStackItem("name" .getBytes(UTF_8))).getString());
```
csmuller, line 260: Remove unintentional space.

csmuller (changes requested):
Good stuff! Requested some changes.

csmuller (approved):

PR #558, opened 2021-05-26
Fix for bug in negative number handling in ScriptBuilder

csmuller:


gsmachado (approved):
Solid! 💪 
Good tests! 💯

PR #569, opened 2021-06-02
Changed method name from 'as' to 'to'

csmuller:
I think it is more appropriate to always use 'to' for such conversion methods.


gsmachado (approved):
💯 Great.

Key updates. 👍

PR #573, opened 2021-06-03
Add toJson method for transaction

csmuller:
For convenience when we want to print the transaction in readable format.


gsmachado (approved):
Good addition for convenience! 👍

Issue #577, opened 2021-06-07
For-loops lead to runtime error when used in objects instantiated by contract class


csmuller:
For-loops lead to runtime errors when they are implemented in classes that are not the main contract class and that are instantiated by the contract.
E.g.: 
```java
    public class MyClass {   
         public void bar() { 
             for (int i = 9; i < 0; i++) {
                  ...
             }
         } 
     } 
```

See hurui200320/neo-n3-playground#2 for more details.

csmuller:
Solved in #579. The fix is available in the latest SNAPSHOT release.

PR #579, opened 2021-06-08
Fix for instance methods bug

csmuller:
Solves #577.
There was a problem with setting up methods correctly when they have the `this` object as a parameter.


gsmachado (approved):
Good catch. 👍  💪

PR #581, opened 2021-06-08
Correct witness scope serialization

csmuller:


gsmachado (approved):
Great! Checked the tests! 💪 👍

PR #587, opened 2021-06-09
Add storage getters that include conversion

### LINKED ISSUE #513: Add Storage getters that automatically convert the return type
The `Storage` class of the devpack provides getters that all return a `ByteString`. It is up to the user to convert that return type to anything else.
We should consider adding getter methods that automatically do a conversion.
E.g.: `public static native int Storage.getInt(StorageContext ctx, BytesString key)`

Such methods would require the appropriate conversion to be added on the method with the `@Instruction` annotation.

mialbu:
Closes #513


mialbu:
This pr is currently blocked, because the compiler is not fully ready to handle `@Instruction` and `@Syscall` annotations.

mialbu:
This pr is no longer blocked. The subtask that was blocking this pr was merged in pr #609.

gsmachado:
It's a pity we didn't merge before the release, but it's fine. We can merge asap, then. I can't review this at the moment, though.

devpack/src/main/java/io/neow3j/devpack/Storage.java
```diff
@@ -46,6 +46,56 @@
...
+     * <p>
+     * This does not incur any extra GAS costs.
+     *
+     * @param context The storage context to search in.
+     * @param key     The key to search for.
+     * @return the value corresponding to the given key as a string.
+     */
+    @Instruction(opcode = OpCode.SWAP)
+    @Instruction(interopService = SYSTEM_STORAGE_GET)
+    @Instruction()
```
csmuller, line 74: It should work without this empty `Instruction`. Did you try?

Ah and then you can actually simply have only one `Instruction` with the interopService attribute set. The SWAP is taken care of the compiler itself.

Same for the other String related methods

devpack/src/main/java/io/neow3j/devpack/Storage.java
```diff
@@ -46,6 +46,56 @@
...
+    public static native String getString(StorageContext context, byte[] key);
+
+    /**
+     * Returns the value corresponding to the given key and converts it to an integer. The bytes
+     * are read in little-endian format. E.g., the byte string {@code 0102} (in hexadecimal
+     * representation) is converted to 513.
+     * <p>
+     * This incurs the GAS cost of converting the {@code ByteString} value to an integer.
```
csmuller, line 82: Please mention that it returns an int value of 0 if the key is not found.

Or maybe we should not return 0 but actually null because 0 is a valid integer value and could be misleading. I mean, just because 0 is returned doesn't necessarily mean that no value was found. I.e., 0 is ambiguous. What do you think?

csmuller (changes requested):
Good stuff, though, I left some comments to discuss.

devpack/src/main/java/io/neow3j/devpack/Storage.java
mialbu: Good input, yes, it works!

devpack/src/main/java/io/neow3j/devpack/Storage.java
mialbu: I changed the instructions to return `null` instead of `0` if the value is `null`. I adapted the according methods in the classes `Helper` (`toInt()`) and `ByteString` (`toInteger()`) as well.

gsmachado (approved):

csmuller (changes requested):

csmuller (approved):

PR #588, opened 2021-06-10
permissions, trusts, groups

csmuller:
Removed Groups annotation, because it is not useful in the compilation process. I added a section to the external docs describing how Groups can be added to the manifest.

Changed permissions to be empty by default.

Also, changed the name of the method for calculating the contract script hash of a contract. I think it fits better like this.


gsmachado (approved):
less code = less problems.

💪

mialbu (approved):
💯

PR #590, opened 2021-06-11
More on trusts and permissions

csmuller:
In blockchain we trust... The same handling of '0x' that applied to permission contract hashes and public keys is needed for trusts.


gsmachado (approved):
Nice!

yeah, `throwIfNotValidContractHashOrPubKey()` makes sure that the `hashOrPubKey` var is not empty or null, so, the `addOrClearHexPrefix()` doesn't need to be concerned with that.

All fine! 💪

PR #593, opened 2021-06-14
Fix for manifest trust serialization

csmuller:


gsmachado (approved):
As always, good tests prove that it works. 💪

PR #597, opened 2021-06-16
Standardize the formatting of the devpack's constant names

csmuller:
I know that java's standard for constants is all upper case, but since the devpack is anyway not used for writting standard code and because i think that the camelcase format has better readability, I wanted to have all of them in camelcase with the first letter being upper case.

Some of the constants where already camelcase. This PR standardizes this throughout the whole devpack.

The javadoc changes are not related but I snuck them into this PR anyway since it's not actual code changes.

### DIFF
--- a/compiler/src/test-integration/java/io/neow3j/compiler/CryptoLibIntegrationTest.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/CryptoLibIntegrationTest.java
@@ -53,12 +53,12 @@ public class CryptoLibIntegrationTest {
         String signature =
                 "a30ded6e19be5573a6f6a5ff37c35d4ae76ff35ab4bee03b5b5bfbbef371f812ff70b5b480462807948a2ffb24dd8771484d9ca5a90333f9e6db69a6c8802a63";
         NeoInvokeFunction response = ct.callInvokeFunction(testName, byteArray(message),
-                publicKey(pubKey), byteArray(signature), integer(NamedCurve.SECP256R1));
+                publicKey(pubKey), byteArray(signature), integer(NamedCurve.secp256r1));
         assertTrue(response.getInvocationResult().getStack().get(0).getBoolean());
 
         message = "0102030406"; // small change in message.
         response = ct.callInvokeFunction(testName, byteArray(message), publicKey(pubKey),
-                byteArray(signature), integer(NamedCurve.SECP256R1));
+                byteArray(signature), integer(NamedCurve.secp256r1));
         assertFalse(response.getInvocationResult().getStack().get(0).getBoolean());
     }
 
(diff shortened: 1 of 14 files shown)


compiler/src/test-integration/java/io/neow3j/compiler/CryptoLibIntegrationTest.java
```diff
@@ -53,12 +53,12 @@ public void verifyWithECDsa() throws IOException {
         String signature =
                 "a30ded6e19be5573a6f6a5ff37c35d4ae76ff35ab4bee03b5b5bfbbef371f812ff70b5b480462807948a2ffb24dd8771484d9ca5a90333f9e6db69a6c8802a63";
         NeoInvokeFunction response = ct.callInvokeFunction(testName, byteArray(message),
-                publicKey(pubKey), byteArray(signature), integer(NamedCurve.SECP256R1));
+                publicKey(pubKey), byteArray(signature), integer(NamedCurve.secp256r1));
```
mialbu, line 56: This should be `Secp256r1` now. The CI did not catch that, since the integration test files are not included in the build.

```suggestion
                publicKey(pubKey), byteArray(signature), integer(NamedCurve.Secp256r1));
```

compiler/src/test-integration/java/io/neow3j/compiler/CryptoLibIntegrationTest.java
```diff
@@ -53,12 +53,12 @@ public void verifyWithECDsa() throws IOException {
...
         assertTrue(response.getInvocationResult().getStack().get(0).getBoolean());
 
         message = "0102030406"; // small change in message.
         response = ct.callInvokeFunction(testName, byteArray(message), publicKey(pubKey),
-                byteArray(signature), integer(NamedCurve.SECP256R1));
+                byteArray(signature), integer(NamedCurve.secp256r1));
```
mialbu, line 61: Here as well.

```suggestion
                byteArray(signature), integer(NamedCurve.Secp256r1));
```

devpack/src/main/java/io/neow3j/devpack/Hash160.java
```diff
@@ -32,6 +34,11 @@
 
     /**
      * Checks if this {@code Hash160} is valid, i.e. is 20 bytes long.
+     * <p>
+     * This method is useful when a contract method takes a {@code Hash160} as a parameter and
+     * you want to make sure that the underlying value really is a valid hash. The contract does
+     * not include or enforce this check automatically, because that would consume extra GAS even
+     * if you don't require that check.
```
mialbu, line 41: To be congruent with the same method for `Hash256`:

```suggestion
     * This method is useful when a contract method takes a {@code Hash160} as a parameter and
     * you want to make sure that the underlying value really is a valid hash. The compiler and
     * NeoVM don't include or enforce this check automatically, because that would consume extra
     * GAS even if you don't require that check.
```

mialbu (changes requested):
Nice! Just requested some minor changes.

devpack/src/main/java/io/neow3j/devpack/Hash160.java
csmuller: Good catch!

mialbu (approved):
:100:

PR #598, opened 2021-06-17
Static variables in debug info

csmuller:


csmuller:
Finally, got this ready. This adds the static variables to the DebugInfo. Additionally, i removed the events from the static variables. I.e. the JVM byte code would treat Events as normal field variables but for the NeoVM script they are only a definition and should not be included in the static variables.

gsmachado:
This is AWESOME! 👍  Looking to it.

PR #599, opened 2021-06-18
Fixes for failed tests on windows

csmuller:


gsmachado (approved):
👍  💪  Test compatibility on other OSs is indeed important! great!

PR #604, opened 2021-07-01
Make Hash160/256.isValid() a static class method

csmuller:
When working on the governance contracts i realized that a static 'isValid' method on the hash classes would be more convenient than the instance methods that we provide at the moment. This new 'isValid(Object data)' checks if the hash is either a Buffer or ByteString and has the right length. Note, that the constructors also do the size check, i.e., there is a bit of redundancy.


mialbu (approved):
:100:

PR #605, opened 2021-07-01
Add NFT classes to devpack

### LINKED ISSUE #474: Add a NonFungibleToken class to the devpack
The neow3j SDK supports a NFT wrapper, but the devpack is lacking that support.
It makes sense to add an NFT class before implementing the NameService contract interface in the devpack since the NameService can be implemented as a subclass of NFT.

mialbu:
Closes #474

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
+++ b/compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
@@ -284,7 +284,7 @@ public class MethodsConverter implements Converter {
                 isContractInterface = true;
                 break;
             }
-            clazz = getAsmClassForInternalName(owner.superName, compUnit.getClassLoader());
+            clazz = getAsmClassForInternalName(clazz.superName, compUnit.getClassLoader());
         }
         if (hasContractHash && !isContractInterface) {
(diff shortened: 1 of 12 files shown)


mialbu:
The integration tests are not running. They are stuck before even running the setUp method and are running forever... I could not yet figure out why. The other integration tests ran fine on my machine.

gsmachado:
> The integration tests are not running. They are stuck before even running the setUp method and are running forever... I could not yet figure out why. The other integration tests ran fine on my machine.

Hmm... the integration tests are not running on your machine? Because it's a known problem that the integration tests are stuck on the nucbox GitHub runner.

mialbu:
All integration tests are running fine, except the ones for the NFT classes. Those are stuck, i.e., they are running but stay at 'Instantiating tests...' forever, so that not even the 'setUp' method is reached. I don't think it's a local problem. If it were, I would assume all other integration tests would also not run fine.

csmuller:
I'm really sorry. There was a bug in the compiler. The compiler didn't deal correctly with the deeper class hierarchy that emerged form the NFT contract interfaces implementation. It works now.

csmuller (approved):

PR #608, opened 2021-07-06
Class compatibility

### LINKED ISSUE #553: Check .class compatibility in the Compiler
Based on the feedback from #546, we don't actually need to enforce 1.8 target compatibility on the `Neow3jCompiler` class -- because that's the responsibility of the Gradle plugin or external tooling.

However, IMHO, we need to simply check Java version of the received .class bytecode. 👍 

If the bytecode we're trying to compile is not 1.8-compatible, raise an early exception with a friendly message. 😄  🚀

gsmachado:
Closes #553 

@csmuller please, take a look at the tests, as well as the changes I've made to the `Compiler.java` file. Thanks!


csmuller (approved):
👍  Good test in the Gradle plugin.

PR #610, opened 2021-07-08
Add neo-express RPC methods

### LINKED ISSUE #606: Add neo-express specific RPC methods to neow3j
Neo-express provides more RPC methods than a normal Neo node (e.g. `expresslistcontracts`, which lists all deployed contracts).
We should consider adding these methods to a separate `Neow3j` class, because they can be useful in the development of developer tools.

mialbu:
Closes #606 #611


core/src/main/java/io/neow3j/protocol/core/Neow3jExpress.java
```diff
@@ -0,0 +1,158 @@
+package io.neow3j.protocol.core;
```
csmuller, line 1: I think this class should reside next to `Neow3j` in the package `io.neow3j.protocol`.

gsmachado: agree

csmuller:
I expeimented a bit with neo-express in the `Neow3jExpressIntegrationTest`s. There are some unused methods there now. The PR is therefore not yet ready to merge.

I think optimally, we would use a neoxp checkpoint file to setup the test environment in the beginning. That way the initialization time is minimal (no waiting for transactions). BUT, the checkpoint file can only be loaded when neoxp is not running. I.e., we need to load it at the startup of the container. Stopping neoxp later is not feasible because the container shuts down immediately (neoxp is its only process).

core/src/test-integration/java/io/neow3j/protocol/Neow3jExpressIntegrationTest.java
```diff
@@ -40,44 +41,41 @@
...
+
     protected static Neow3jExpress neow3jExpress;
 
     @ClassRule
-    public static NeoTestContainer neoTestContainer = neoExpressTestContainer(1);
+    public static NeoExpressTestContainer neoTestContainer = new NeoExpressTestContainer(1,
+            NEF_FILE_SOURCE, NEF_FILE_DESTINATION,
+            MANIFEST_FILE_SOURCE, MANIFEST_FILE_DESTINATION,
+            INVOKE_FILE_SOURCE, INVOKE_FILE_DESTINATION
+    );
```
mialbu, line 63: I couldn't find this new class `NeoExpressTestContainer`. Please push it as soon as it's ready. :+1:

csmuller (approved):

csmuller (approved):

PR #615, opened 2021-07-13
Neow3j setter in Transaction class

### LINKED ISSUE #614: Add setter for Neow3j object in Transaction class
**Problem:**

When exporting the transaction bytes (here referred to as `txBytes`) and reimporting them with `NeoSerializableInterface.from(txBytes, Transaction.class)`, the just created `Transaction` object cannot be used because its `Neow3j` object is `null` and there is no option to set it.

**Solution suggestion:**

Add a setter for the `Neow3j` object in the class `Transaction`.

mialbu:
Closes #614


csmuller (approved):
All good!

csmuller (approved):

PR #620, opened 2021-07-27
Gradle: change compile to implementation

### LINKED ISSUE #485: Change dependencies from 'compile' to 'implementation'
The `compile` directive that we use in all our `build.gradle` files is depricated. We should replace it with `implementation`. Equally for the `compileTest` that should be replaced by `implementationTest`.

The advantage of `implementation` is that it avoids that dependencies of our library are transitively passed on to the user. So if the dev depends on `io.neow3j:devpack`, the dependencies of the devpack do not show up in the dev's classpath. Or, if someone needs the `contract` and the `core` package, they have to depend on both explicitly.

This includes changing the README.md in this repo and the external documentation, where we usually say that you only need to import `contract` to also get `core`.

gsmachado:
Closes #485

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -130,7 +130,7 @@ subprojects {
     }
 
     dependencies {
-        testCompile "junit:junit:$junitVersion",
+        testImplementation "junit:junit:$junitVersion",
                 "org.mockito:mockito-core:$mockitoVersion",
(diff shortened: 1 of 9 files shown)


gsmachado:
`./gradlew clean test` is not yet compiling.

csmuller (approved):

PR #621, opened 2021-07-30
Change debug info default column numbers to 1

csmuller:
VSCode column numbers and lines start at 1, thus the default column number should be set to 1 and not 0.


gsmachado (approved):
All fine from my point of view!

mialbu (approved):
:+1:

Issue #601, opened 2021-06-27
Static variables from contract class don't work correctly in auxiliary classes


hurui200320:
Hi, I'd like to move some helper functions (in contract class which make it longer than 300+ lines) to a separate class so it will look neat and clean. Since those helper functions are specific to that contract class, it will be great if I can just use that static field rather than pass it as a parameter.

Issue: When using `StorageMap` as a static field, it gives `Invalid type for PICKITEM: InteropInterface` error.
Code:
```java
public class TestContract {
    private static final StorageContext CTX = Storage.getStorageContext();
    static StorageMap MAP = CTX.createMap("some_prefix");

    public static void foo(){
        Auxiliary.helper();
    }
}

public class Auxiliary {
    static void helper() {
        TestContract.MAP.get("key");
    }
}
```

hurui200320:
Also a weird thing about static field in helper class.
There are two static field in contract class:
```java
public class TestContract {
    // 0x218bc33597ce7dc32079f593d5ed5766c1ef7764
    static final Hash160 HASH1 = addressToScriptHash("NV5CSGyT6B39fZJ6zw4x6gh1b3C6cpjTm3");
    // 0xf9077f4eec533d175eed50b789b4d8d55dacf5ff
    static final Hash160 HASH2 = addressToScriptHash("NjFMoMSoukNBetDZYPsGKzpLrUA1zgkMNM");

    public static Hash160 foo(){
        return Auxiliary.helper();
    }
}
```
Then no matter I use first or second one, it always gives me the first one:
```java
public class Auxiliary {
    static Hash160 helper() {
//        return TestContract.HASH1;
        return TestContract.HASH2;
    }
}
```
`foo` return: `6477efc16657edd593f57920c37dce9735c38b21`, which is the little endian representation of the HASH1

gsmachado:
as far as I understood by talking to @csmuller, this was partially solved in #624. 🚀

csmuller:
Sorry, i mixed two issues up. This one is solved. Issue #616 is the one that has only partially been tackled by #624.

PR #624, opened 2021-08-06
Correct handling of static field variables

csmuller:


csmuller:
I originally wanted to make it possible to declare static variables in non-contract classes as well. But, that implies a lot of changes because the variable declaration has to be merged into the INITSSLOT method of the NeoVM script. I.e., we would have multiple `<clinit>` methods (static constructors) from the JVM bytecode that we have to merge into one INITSSLOT method in the NeoVM code. The compiler cannot handle this well at the moment. Therefore, I went for the simpler approach which rejects static variable declaration in non-contract classes.
Still, on the way I introduced some changes that go into the direction of allowing static variables in non-contract classes. I didn't revert those changes because they don't change the compilers behaviour and might come in handy at a later time.

gsmachado:
> I originally wanted to make it possible to declare static variables in non-contract classes as well. But, that implies a lot of changes because the variable declaration has to be merged into the INITSSLOT method of the NeoVM script. I.e., we would have multiple `<clinit>` methods (static constructors) from the JVM bytecode that we have to merge into one INITSSLOT method in the NeoVM code. The compiler cannot handle this well at the moment. Therefore, I went for the simpler approach which rejects static variable declaration in non-contract classes.
> Still, on the way I introduced some changes that go into the direction of allowing static variables in non-contract classes. I didn't revert those changes because they don't change the compilers behaviour and might come in handy at a later time.

Do we have an issue to handle this in future? If not, can we create one?

csmuller:
Issue #601 covers this. I will not close it once this PR is merged, but update its description a bit.

gsmachado (approved):

PR #627, opened 2021-08-11
TransactionBuilder rework

### LINKED ISSUE #603: Rework the TransactionBuilder by removing the Wallet property
We should consider removing the wallet from the `TransactionBuilder`.  It makes the builder less intuitive. When developers add a signer to the builder and provide the signers account, why should they also have to pass the wallet when the account already contains the private key.

The only functionality that we loose is the automatic signing with a multi-sig address if all participating accounts are in the wallet. But, this is probably not a real-world use case anyway. Having all private keys of the multi-sig address in one wallet partly defeats the use of the multi-sig.

mialbu:
Closes #603


csmuller:
image

csmuller (changes requested):
The `signerHash` field variable in `Signer` should now be moved up to `ContractSigner` because the `AccountSigner` does not need that variable anymore.
Basically, this is about making `getScriptHash()` abstract in `Signer` and implementing it in `AccountSigner` and `ContractSigner`. Then using it in `Signer` wherever `signerHash` is used.

core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
```diff
@@ -468,14 +441,12 @@ public Transaction sign() throws Throwable {
...
-                // There's no need to check if every signer has its account in the wallet here.
-                // This check has already been executed within building the transaction above
-                // when calculating the network fees.
-                Account signerAcc = wallet.getAccount(signer.getScriptHash());
-                if (signerAcc.isMultiSig()) {
-                    signWithMultiSigAccount(txBytes, signerAcc);
+                Account a = ((AccountSigner) signer).getAccount();
+                if (a.isMultiSig()) {
+                    throw new IllegalStateException("Transactions with multi-sig signers cannot " +
+                            "be signed automatically.");
```
csmuller, line 447: Please add a test for this exception.

core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
```diff
@@ -389,19 +363,18 @@ private long getSystemFeeForScript() throws IOException {
     private long calcNetworkFee() throws IOException {
```
csmuller, line 363: Two exceptions thrown in this method are out of date and still contain references to a wallet. Please adapt them. 
(I couldn't mark the code locations because there not affected by any existing changes.)

csmuller (changes requested):

core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
mialbu: done

csmuller (approved):
💯

PR #628, opened 2021-08-12
Add manifest structure to devpack

csmuller:


gsmachado:
@csmuller so, as far as I could see, this PR just transforms the `String` manifest into the `Manifest` objects and its subsequent children objects, right?

If that's the case, does it make sense to have serialization/deserialization tests?

csmuller:
The implementation in the core changed such that the manifest is not a string anymore. The new `Manifest` class represents the correct structure in which the returned value comes.
It's a change that we missed in the past.

gsmachado (approved):
💪

PR #630, opened 2021-08-16
Fix for serialisation of manifest permissions

csmuller:


gsmachado (approved):
Good work. Now we properly have tests for the corner cases, in my point of view. 💪

PR #634, opened 2021-08-30
Add transfer methods with hash instead of account as 'from' parameter

csmuller:
While adapting the examples I realised the need for transfer methods that take a Hash160 as the 'from' parameter. That allows devs to create transfers from ContractSigners.


gsmachado (approved):
even better with adapted tests. 👍 💪 😄

PR #637, opened 2021-09-09
Add cancel vote method

csmuller:


contract/src/main/java/io/neow3j/contract/NeoToken.java
```diff
@@ -268,15 +269,42 @@ public TransactionBuilder vote(Account voter, ECPublicKey candidate) throws IOEx
      * @throws IOException if there was a problem fetching information from the Neo node.
      */
     public TransactionBuilder vote(Hash160 voter, ECPublicKey candidate) throws IOException {
-        if (!isCandidate(candidate)) {
-            throw new IllegalArgumentException("The provided public key is not a candidate. Only " +
-                    "candidates can be voted for.");
+        if (candidate == null) {
```
gsmachado, line 272: Can you add in the Java docs something like this?

"If the candidate is set to null, then the vote is withdrawn. See .cancelVote()."

gsmachado (approved):
Just added a comment about docs.

Implementation-wise it's ready to go! 💪

contract/src/main/java/io/neow3j/contract/NeoToken.java
csmuller: 👍 Done.

PR #644, opened 2021-09-21
Test module

csmuller:


compiler/build.gradle
```diff
@@ -11,5 +11,6 @@ dependencies {
 
     testImplementation "ch.qos.logback:logback-classic:$logbackVersion",
             "com.github.tomakehurst:wiremock-jre8:$wiremockVersion",
+            "org.testcontainers:testcontainers:$testcontainersVersion",
```
gsmachado, line 14: why are you adding the testcontainers dependencies to all the projects?

core/src/main/java/io/neow3j/crypto/ECKeyPair.java
```diff
@@ -207,17 +205,7 @@ private static KeyPair createSecp256r1KeyPair() throws NoSuchProviderException,
...
-                toBytesPadded(getPrivateKey().getInt(), NeoConstants.PRIVATE_KEY_SIZE),
-                new byte[]{(byte) 0x01}
-        );
-        byte[] checksum = hash256(data, 0, data.length);
-        byte[] first4Bytes = Arrays.copyOfRange(checksum, 0, 4);
-        data = concatenate(data, first4Bytes);
-        String wif = Base58.encode(data);
-        Arrays.fill(data, (byte) 0);
-        return wif;
+        return WIF.getWIFFromPrivateKey(getPrivateKey().getBytes());
```
gsmachado, line 208: was this redundant code?

neow3j-test/build.gradle
```diff
@@ -0,0 +1,18 @@
+description 'test tools for neow3j'
```
gsmachado, line 1:
```suggestion
description 'test framework for neow3j'
```

or do you prefer to leave as "tools"? no strong feelings here.

neow3j-test/src/main/java/io/neow3j/test/ContractTestExtension.java
```diff
@@ -0,0 +1,213 @@
...
+//    private final static String ACC_1_ADDR = ACC_PROPS_PREFIX + "1.address";
+//    private final static String ACC_1_SCRIPT_HASH = ACC_PROPS_PREFIX + "1.script-hash";
+//    private final static String ACC_1_PRIV_KEY = ACC_PROPS_PREFIX + "1.private-key";
+
+    //    private Properties properties;
+    private NeoExpressTestContainer container;
+    private Neow3jExpress neow3j;
+    private SmartContract contractUnderTest;
+
+//    public ContractTestExtension() {
```
gsmachado, line 61: oh.... now I get what you were trying to do here.... so, basically, you wanted to "read" files from resources where a certain prefix/suffix is set, and then inject those to the neo-express. Is that right?

In addition to that, I think the test framework should rely on an existing neo-express config file, if specified by the user... but maybe this is something up to discussion in the future.

neow3j-test/src/test/java/io/neow3j/test/ExampleTest.java
```diff
@@ -0,0 +1,53 @@
...
+import io.neow3j.wallet.Account;
+import org.junit.Ignore;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import java.math.BigInteger;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Ignore
```
gsmachado, line 18: ahh... I get it... this annotation is used just to tell JUnit to exclude the tests from this class (if executed from gradle). Or is it used for something else?

test-tools/src/main/java/io/neow3j/test/NeoExpressTestContainer.java
```diff
@@ -1,64 +1,192 @@
...
-     * Default location for a neoxp checkpoint file in the container.
-     */
-    public static final String CHECKPOINT_DEST = "/app/setup.neoxp-checkpoint";
+import java.math.BigInteger;
 
-    /**
-     * Default location for a neoxp batch file in the container.
-     */
-    public static final String BATCH_DEST = "/app/setup.batch";
+public class NeoExpressTestContainer extends GenericContainer<NeoExpressTestContainer> {
```
gsmachado, line 10: Can we make an interface from this as we discussed in a call? To enable extensibility in the future.

Then, the implementations could be: `NeoExpressTestDocker` and `NeoExpressTestLocal` or something like that. Wdyt?

gsmachado (changes requested):
Absolutely great work! 💪 

Just take a look at my comments before I approve. 👍 💯

compiler/build.gradle
csmuller: I removed it from the base build.gradle because it is not needed in all modules and moved it into the ones that actually require it. Just neuroticism at work.

core/src/main/java/io/neow3j/crypto/ECKeyPair.java
csmuller: Yes.

neow3j-test/src/main/java/io/neow3j/test/ContractTestExtension.java
csmuller: First, I wanted to use a properties file where we would set the default account information (e.g., private keys) that are included in the default neo-express config. Later, I changed to the approach of actually reading the account information from the config JSON. Then, this properties information became obsolete. I left it here in case we might make use of properties in the future. Though, I can remove it now since it is available in at least one commit.

csmuller: The user can define her own neoxp config that will overwrite the default one.

test-tools/src/main/java/io/neow3j/test/NeoExpressTestContainer.java
csmuller: Yes, that makes sense. But do you think it is necessary to create the interface now? I prefer doing it only as soon as we have at least two implementations.

neow3j-test/src/test/java/io/neow3j/test/ExampleTest.java
csmuller: I only added this because this test was still in unit tests even though it's an integration test. I moved it to an integrationTest folder now and removed the `@Ignore`

compiler/build.gradle
gsmachado: All fine!

gsmachado (approved):
💪 👍 💯

PR #649, opened 2021-10-08
Signature related improvements

### LINKED ISSUE #648: Update CryptoLib and signature contract parameter
The method `verifyWithECDsa` in the native contract `CryptoLib` should take a `ByteString` instead of a `String` for the `signature` parameter. Further, the contract parameter of type `signature` should accept the type `SignatureData` as parameter, such that a signature can easily be created and passed as parameter with ease.

**Tasks**
- [x] Update `verifyWithECDsa`
- [x] Accept `SignatureData` in signature contract parameter

mialbu:
Closes #648

### DIFF
--- a/core/src/test/java/io/neow3j/crypto/SignTest.java
+++ b/core/src/test/java/io/neow3j/crypto/SignTest.java
@@ -2,7 +2,9 @@ package io.neow3j.crypto;
 
 import static io.neow3j.crypto.Hash.sha256;
 import static io.neow3j.crypto.Sign.recoverSigningScriptHash;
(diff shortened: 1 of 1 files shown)


csmuller:
Test coverage must not decrease! 😈

gsmachado:
@mialbu yeah, I realized that coverage decreased due to lines not even included in this PR. That's why I proposed to review the config and settings of our coverage, somehow.

So, if you guys think it's fine, let's move ahead and merge even with the slight decrease on the coverage. Let's just keep this in mind that we do our best for it. 👍🙏

PR #655, opened 2021-10-13
Release 3.13.1

csmuller:


csmuller:
Codecov is not updating the above statistics. With the last few changes the code coverage should not be decreasing with this PR.

PR #669, opened 2021-11-01
Support getMessage method for caught exception

### LINKED ISSUE #662: Support getMessage method on Exceptions

When you catch an exception in a smart contract and want to get the message string with the `getMessage` method, the contract will fail at runtime. On the NeoVM the java `Exception` object is just the string that has been passed to the `Exception` constructor (or "error" when no string was provided). Thus, when calling `getMessage` on that string, the NeoVM fails.

Adapt the neow3j compiler such that calls to `Exception.getMessage()` are handled specially. In theory the call to `getMessage()` simply has to be ignored since the exception string is already at the top of the stack.

Tasks:
- [x] Adapt the compiler
- [x] Write integration tests in which exceptions are thrown (e.g., by assertions) and then the `Exception` object is handled in the `catch` clause by calling its `getMessage()` method.
- [x] Update the section Smart Contract Development -> NeowJava -> exceptions in the neow3j-docs.

mialbu:
Resolves #662

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
+++ b/compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
@@ -100,6 +100,7 @@ public class MethodsConverter implements Converter {
     private static final String EQUALS_METHOD_NAME = "equals";
     private static final String LENGTH_METHOD_NAME = "length";
     private static final String GET_CONTRACT_HASH_METHOD_NAME = "getHash";
+    private static final String THROWABLE_GET_MESSAGE = "getMessage";
 
 
     /**
@@ -183,6 +184,9 @@ public class MethodsConverter implements Converter {
             callingNeoMethod.addInstruction(new NeoInstruction(OpCode.SIZE));
             return methodInsn;
         }
+        if (isThrowableGetMessage(owner, methodInsn)) {
+            return methodInsn;
+        }
         NeoMethod calledNeoMethod = new NeoMethod(calledAsmMethod, owner);
         compUnit.getNeoModule().addMethod(calledNeoMethod);
         calledNeoMethod.initialize(compUnit);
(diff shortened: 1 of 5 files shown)


gsmachado:
why the int tests are failing?

It seems related to some kind of problem with cache, that @csmuller changed recently.

I will trigger it again just to see if the problem vanishes.

mialbu:
Yes, it's some cache issue in the `Set up JDK 1.8` task. @csmuller is looking into it.

mialbu:
Btw, the tests passed in the commit before and the latest commit only contains code style fixes.

compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
```diff
@@ -194,6 +197,12 @@ private static AbstractInsnNode handleUncachedMethodCall(NeoMethod callingNeoMet
         return methodInsn;
     }
 
+    private static boolean isThrowableGetMessage(ClassNode owner, MethodInsnNode methodInsn) {
+        return getFullyQualifiedNameForInternalName(owner.name).equals(
+                Throwable.class.getCanonicalName()) &&
+                methodInsn.name.equals(THROWABLE_GET_MESSAGE);
```
csmuller, line 203: Wait, so the owner is `Throwable` and not `Exception` or `AssertionError`? That's convenient.

csmuller (approved):
👍

compiler/src/main/java/io/neow3j/compiler/converters/MethodsConverter.java
mialbu: Yes, both `AssertionError` and `Exception` use the method `getMessage` from their shared super class `Throwable`, which is super convenient. 🙂

PR #670, opened 2021-11-01
Support integer keys in storage methods

### LINKED ISSUE #667: Add Storage methods that take int as the key

In smart contracts we usually have many storage keys that are setup as static variables on the contract. These keys can have small values that fit into a byte and naturally one wants to express them with a byte or an integer. But our Storage methods don't allow those types as keys. Thus, the developer has to either define them as `new byte[]{key}` or create a new byte array every time the key is used. Both incur higher costs compared to just having a byte or int. In the first case the java compiler does not inline the `new byte[]{key}` but actually creates a static variable that appears in the contracts INITSSLOT method. With an int or byte the key would be inlined. With the second case there is an obvious over head of creating a new buffer every time the key is used.

Create equivalents to all Storage and StorageMap methods but with an `int` as the key. It's enough to do this with `int`, because a `byte` parameter can be passed to those methods as well.

mialbu:
Closes #667


csmuller (approved):
🦾

PR #684, opened 2021-11-18
Attempt to decouple test framework from neo-express

csmuller:


test-tools/src/main/java/io/neow3j/test/NeoTestContainer.java
```diff
@@ -72,12 +72,69 @@ public NeoTestContainer(String configFileSource) {
...
+    }
+
+    @Override
+    public TestBlockchain withCheckpoint(String checkpointFile) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public TestBlockchain withConfigFile(String configFile) {
+        return null;
```
gsmachado, line 92: So, all the methods will `null` here are, in fact, not implemented, right?

If so, can you throw `java.lang.UnsupportedOperationException`?

gsmachado: or, maybe it's supported, but it's not implemented by us yet?

In this case, I would throw another exception.

test-tools/src/main/java/io/neow3j/test/TestBlockchain.java
```diff
@@ -0,0 +1,157 @@
+package io.neow3j.test;
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+public interface TestBlockchain {
```
gsmachado, line 7: Why don't you make `TestBlockchain<T>` and then in each method that returns `TestBlockchain` (e.g., `withSecondsPerBlock`) you simply do `T withSecondsPerBlock(int secondsPerBlock);`

I believe that's possible, but you need to check. 😄

gsmachado (commented):
Just some comments. 👍 

Good work, BTW. 💪

test-tools/src/main/java/io/neow3j/test/NeoTestContainer.java
csmuller: Ouh, it was not my intention to push the changes on `NeoTestContainer` 😟 
We should have a call about this, deciding if it even makes sense to have `NeoTestContainer` implement `TestBlockchain`.

test-tools/src/main/java/io/neow3j/test/TestBlockchain.java
csmuller: hmmm, I don't see the rationale behind this. 
Then, if you used an instance of `TestBlockchain` as follows, you get back type `T` which isn't a `TestBlockchain` anymore, i.e., method chaining is not possible.
```
chain.withSecondsPerBlock(annotation.blockTime()). ?
```

gsmachado: hmmm.. the `chain` will be a `TestBlockchain` anyway... because the class of the `chain` variable will implement the interface anyway.

Also, it will give the possibility to developers to chain other methods they would like to implement (besides the TestBlockchain interface).

gsmachado (approved):
Good to go! 😄 👍  

Great work.

PR #687, opened 2021-11-19
WitnessRules

csmuller:


gsmachado:
Awesome. Do we already have an issue tracking all the new changes for the upcoming neo-node version?

csmuller:
> Awesome. Do we already have an issue tracking all the new changes for the upcoming neo-node version?

We have the label "Neo 3.1.0". All necessary changes are tagged with that.

csmuller:
Waiting for Neo 3.1.0 release and update of our test node.

csmuller:
This feature has to wait till Neo supports WitnessRules in the `invokescript` and `invokefunction` RPC methods. Till then neow3j will not be able to use WitnessRules because fetching the system fee fails everytime.

PR #705, opened 2022-01-12
Correct multi-sig script creation

csmuller:


core/src/main/java/io/neow3j/crypto/ECKeyPair.java
```diff
@@ -316,7 +316,7 @@ public int hashCode() {
 
     }
 
-    public static class ECPublicKey extends NeoSerializable {
+    public static class ECPublicKey extends NeoSerializable implements Comparable<ECPublicKey> {
```
gsmachado, line 319: 😍

core/src/main/java/io/neow3j/crypto/ECKeyPair.java
```diff
@@ -441,5 +441,15 @@ public boolean equals(Object o) {
         public int hashCode() {
             return ecPoint.hashCode();
         }
+
+        @Override
+        public int compareTo(ECPublicKey o) {
```
gsmachado, line 446: 😍

mialbu (approved):
:100:

PR #706, opened 2022-01-13
Improvement of storage-related classes

### LINKED ISSUE #559: Improve storage-related classes
There are several improvments we can make for the devpack's storage classes:
- [x] Remove the `createMap(...)` methods from `StorageContext` and add corresponding constructors to the `StorageMap` class. This removes an intermediate method and method call from the contract script. It's also more intuitive to use the constructor in my opinion.
- [x] Replace all method bodies in `StorageMap` with annotations exclusively. This removes another method and call from the script and replaces it with inline instructions.
  - [x] This requires that `@Instruction` and `@Syscall` annotations can be used at the same time on a method.
- [x] Add `getBoolean` methods to `Storage` and `StorageMap`

mialbu:
Resolves #559

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,9 +15,9 @@ buildscript {
     ext.jetbrainsAnnotationsVersion = '20.1.0'
     ext.mockitoVersion = '3.6.0'
     ext.hamcrestVersion = '1.3'
-    ext.testcontainersVersion = '1.15.3'
+    ext.testcontainersVersion = '1.16.2'
     ext.awaitility = '3.1.5'
-    ext.wiremockVersion = '2.25.1'
+    ext.wiremockVersion = '2.32.0'
     ext.jupiterVersion = '5.8.0'
 
     repositories {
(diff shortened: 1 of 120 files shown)


csmuller (changes requested):
We have an issue with some storage tests that i don't think test exactly what is intended. The problem is that, in some tests, the stack item types created on the NeoVM are not the actua types that the test intends.
I added some `assert`s in some tests to check that. Thus, some tests fail now. Check it out
Example:
        public static boolean getBooleanByByteArrayKey(byte[] key) {
            assert key instanceof byte[];
            return Storage.getBoolean(ctx, key);
        }
```
So, please make sure that the correct stack item type is used in each test. I know this might be an issue for older tests as well, but don't touch those.

Furthermore, I looked at the new `StorageMap` constructors and found room for optimisation by using `@Instruction` instead of a method body. I added two TODOs on two constructors to finish this work. Checkout the neo-devpack-dotnet as a reference on how to do this.

csmuller:
💪 good work!

PR #708, opened 2022-01-21
Allow empty array contract parameter

csmuller:


core/src/test/java/io/neow3j/contract/ContractParameterTest.java
```diff
@@ -160,24 +161,18 @@ public void testArrayParamCreationFromObjects_failUnSupportedObject() {
...
 
     @Test
-    public void testArrayParam_Empty() {
-        exceptionRule.expect(IllegalArgumentException.class);
-        exceptionRule.expectMessage("At least one parameter is required");
-        array();
+    public void testArrayParam_Null() {
+        ContractParameter param = array((Object) null);
+        ContractParameter[] value = (ContractParameter[]) param.getValue();
+        assertThat(value.length, is(1));
```
gsmachado, line 174: hmmm, I don't get it... why is this `1`?

A reference to a `null` Object is also in the array/stack?

csmuller: I was also confused first. But `array((Object) null)` means that we'll actually have an array with one element and that element is of parameter type `ANY` with value `null`.

gsmachado: hmmmmmm... not that intuitive, to be honest.

Can we document this somewhere? Maybe in a section called "types" (or similar) in neow3j.io website?

csmuller: I think it's rather weird to do what i did in this test.
The intuitive way of getting an empty array would be `array()`, which works well. So i don't think there's a problem with the way the method works.
Anyway, I will extend the javadoc on `array(...)`.

gsmachado: great!

gsmachado (approved):
💪  indeed this makes dev's life easier.

PR #710, opened 2022-01-25
Replace ExpectedException with Assert.assertThrows

mialbu:
Refactored tests to replace the deprecated ExpectedException with Assert.assertThrows.


csmuller (approved):
🥵 Extensive! 💪

PR #713, opened 2022-01-27
Support nested Map and List Objects in ContractParameter

### LINKED ISSUE #700: Extend ContractParameter.mapToContractParameter(...) to handle nested Lists and Maps
When creating an Array or Map contract parameter one can pass a generic Java Map or List and the class will try to derive the correct Contract Parameter type for its contents. The corresponding method does not yet handle Maps and Lists within Maps and Lists, i.e., no recursion. Add such functionality.

- [x] Add functionality to handle the conversion of nested Maps and Lists to a ContractParameter.

Additionally
- [x] Add ECPublicKey to the handled objects and look for other types that we could add to the `mapToContractParameter` method.
- [x] Add a `ContractParameter.publicKey` static method that takes an `ECPublicKey`.

mialbu:
Resolves #700.

- Support automatic conversion of nested maps and lists to the corresponding ContractParameter types.
- Support automatic conversion of ECPublicKey and Sign.SignatureData to the corresponding ContractParameter using `mapToContractParameter`.
- Added static method ContractParameter.publicKey(ECPublicKey).

### DIFF
--- a/core/src/main/java/io/neow3j/types/ContractParameter.java
+++ b/core/src/main/java/io/neow3j/types/ContractParameter.java
@@ -198,6 +198,8 @@ public class ContractParameter {
             return hash160((Account) o);
         } else if (o == null) {
             return any(null);
+        } else if (o instanceof List) {
+            return array((List<?>) o);
         } else {
             throw new IllegalArgumentException("The provided object could not be casted into " +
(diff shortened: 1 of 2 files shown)


csmuller (approved):
img

PR #716, opened 2022-02-10
Add NeoStandards enum for SupportedStandards annotation

### LINKED ISSUE #409: Restrict usable values in the SupportedStandards annotation
Currently the annotation `SupportedStandards` takes any string as a valid input. Only actual standards should be accepted in this annotation. This also creates consistency in that standards are written with a dash and with capital letters (as it is currently done in the native contracts).

mialbu:
Resolves #409


mialbu:
> It's still a draft. Should I review already?

It's ready for review. I've opened the pr before we talked about the option to create separate annotations for standards (e.g., `@NEP17`). As discussed, we can use the solution as in this pr. In the future, we can still discuss again if we want to add separate annotations for standards.

devpack/src/main/java/io/neow3j/devpack/annotations/SupportedStandard.java
```diff
@@ -0,0 +1,29 @@
...
+import io.neow3j.devpack.annotations.SupportedStandard.SupportedStandards;
+import io.neow3j.devpack.constants.NeoStandard;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Target;
+
+/**
+ * Used to define features of a smart contract. The attributes will be written to the
+ * contract's manifest file.
```
csmuller, line 12: Clarify in the javadoc that only either the `neoStandard` or the `customStandard` attribute should be set but not both.

csmuller (changes requested):

PR #724, opened 2022-02-18
Add additional system fee method to TransactionBuilder

### LINKED ISSUE #723: add an `addiontalSystemFee` method on `TransactionBuilder`
Because blockchain state could change between creating a transaction and sending it, the actual GAS costs of an invocation could be off from the system fee set on the transaction. 
Therefore, we need a method on the `TransactionBuilder` similar to `additionalNetworkFee` that allows the user to add a higher  system fee than calculated by the node.

mialbu:
Resolves #723

### DIFF
--- a/core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
+++ b/core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
@@ -204,6 +204,9 @@ public class TransactionBuilder {
      * Configures the transaction with an additional system fee.
      * <p>
      * The basic system fee required to send this transaction is added automatically.
+     * <p>
+     * Use this, if you expect the transaction to consume more GAS because of chain state changes happening between
+     * creating the transaction and actually sending it.
      *
      * @param fee the additional system fee in fractions of GAS.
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
```diff
@@ -207,8 +201,21 @@ public TransactionBuilder additionalNetworkFee(long fee) {
     }
 
     /**
-     * Sets the script for this transaction. It defines the actions that this transaction will
-     * perform on the blockchain.
+     * Configures the transaction with an additional system fee.
+     * <p>
+     * The basic system fee required to send this transaction is added automatically.
+     *
```
csmuller, line 207:
```suggestion
     * The basic system fee required to send this transaction is added automatically.
     * <p>
     * Use this, if you expect the transaction to consume more GAS because of chain state changes
     * happening between creating the transaction and actually sending it.
     *
```

csmuller (approved):
Great!
Just added a bit to the javadocs.

PR #730, opened 2022-03-04
Additions to neow3j's Transaction class

csmuller:


csmuller:
Includes:
- Addition of generating a ContractParameterContext object from a transaction (for use in neo-cli)
- Change of variable and getter names in ContractParameter. I'm including this here, because our next release will have breaking changes anyway.
- Removed the '0x' prefix in the `getEncodedCompressedHex` method because neow3j in general doesn't add the hex prefix.

mialbu (approved):
:100:

PR #731, opened 2022-03-09
Ensure correct ordering of signatures

### LINKED ISSUE #728: Ensure correct ordering of signatures in `Transaction.addMultiSigWitness` methods
To make it less error-prone for users to add multi-sig witnesses to a transaction we should adapt the `addMultiSigWitness` methods on `Transaction`.

- [x] `addMultiSigWitness(VerificationScript verificationScript, Account... accounts)` should sort the accounts according to their public keys.
- [x] `addMultiSigWitness(VerificationScript verificationScript, Sign.SignatureData... signatures)` needs a different signature that allows passing a map of public keys to the corresponding signature data. The method should then add the signatures according to their public keys.

mialbu:
Resolves #728


csmuller (approved):
💯 I directly committed changes to the points I noticed.

PR #735, opened 2022-03-16
Fix for isEvent method

### LINKED ISSUE #732: Limited parameters usage in Event fire method
**Description of the bug**
When firing an event with a parameter that contains a method call instruction of a method in neow3j's devpack, the compiler runs into an `UnsatisfiedLinkError`.

**To Reproduce**
The following class can be compiled correctly, since the `value` is a native Java `int`:

```java
public class FireEventWithNativeJavaValue {
        public static int value = 12;

        @DisplayName("onCall")
        public static Event1Arg<Integer> onCall;

        public static void fireEvent() {
            onCall.fire(value);
        }
}
```

When changing the `value` to use the method `StringLiteralHelper.stringToInt()` to derive the `value`, the contract results in a `UnsatisfiedLinkError`.

```java
    static class FireEventWithDevpackMethodCall {
        public static int value = StringLiteralHelper.stringToInt("12");

        @DisplayName("onCall")
        public static Event1Arg<Integer> onCall;

        public static void fireEvent() {
            onCall.fire(value);
        }
}
```

csmuller:
Close #732


compiler/src/main/java/io/neow3j/compiler/InitsslotNeoMethod.java
```diff
@@ -70,17 +70,22 @@ public void convert(CompilationUnit compUnit) throws IOException {
...
     }
 
-    private void throwOnEventConstructorCall(AbstractInsnNode insn) {
+    private void throwOnEventConstructorCall(AbstractInsnNode insn, CompilationUnit compUnit) {
         if (insn instanceof TypeInsnNode) {
             TypeInsnNode typeInsn = (TypeInsnNode) insn;
-            if (isEvent(typeInsn.desc)) {
+            String desc = typeInsn.desc;
+            if (desc.contains("/") && !desc.startsWith("[")) {
+                // In some descriptors AMS doesn't add the "L" and ";" when the type is a class.
```
gsmachado, line 85:
```suggestion
                // In some descriptors ASM doesn't add the "L" and ";" when the type is a class.
```

compiler/src/main/java/io/neow3j/compiler/Compiler.java
```diff
@@ -803,23 +803,19 @@ public static NeoInstruction buildPushNumberInstruction(BigInteger number) {
...
-     */
-    public static boolean isEvent(String classDesc) {
-        String fqn = getFullyQualifiedNameForInternalName(
-                AsmHelper.stripObjectDescriptor(classDesc));
-        Class<?> clazz;
-        try {
-            clazz = Class.forName(fqn);
-        } catch (ClassNotFoundException e) {
+    public static boolean isEvent(String classDesc, CompilationUnit compUnit) {
+        char firstChar = classDesc.charAt(0);
```
gsmachado, line 807: Is there any chance that the `classDesc` is `null` or `.charAt(0)` returns something absurd?

gsmachado: Is there a better way to avoid unexpected exceptions?

csmuller: I think it is rather up to the caller of this method to make sure it doesn't pass null. I.e., if the caller passes null, then the caller has to be fixed, and this method should not just handle the exception and continue gracefully.

csmuller (commented):
Forgot to fix the javadocs.

mialbu (approved):
:100:

PR #738, opened 2022-03-23
Support instance inheritance

### LINKED ISSUE #575: Support inheritance for instance variables
Add support for inheritance of instance variables as used in the following example.

```java

public class MyContract {
    
    static Event2Args<Integer, String> event;

    public static void method(MySubclass obj) {
        event.fire(obj.i, obj.s);
    }

    public class MyClass {
        public int i;
    }


    public class MySubclass extends MyClass {
        public String s;
    }
}
```

mialbu:
Closes #575.

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
+++ b/compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
@@ -385,6 +385,14 @@ public class ObjectsConverter implements Converter {
             calledNeoMethod.convert(compUnit);
             fieldSize = calculateFieldSize(structClassNode, compUnit);
         }
+        return finalizeConstructorCall(fieldSize, typeInsnNode, ctorMethod, structClassNode, callingNeoMethod,
+                calledNeoMethod, compUnit);
+    }
+
+    private static AbstractInsnNode finalizeConstructorCall(int fieldSize, TypeInsnNode typeInsnNode,
+            MethodNode ctorMethod, ClassNode structClassNode, NeoMethod callingNeoMethod, NeoMethod calledNeoMethod,
+            CompilationUnit compUnit) throws IOException {
+
         addPushNumber(fieldSize, callingNeoMethod);
         callingNeoMethod.addInstruction(new NeoInstruction(OpCode.NEWARRAY));
         // TODO: Determine when to use NEWSTRUCT.
@@ -403,6 +411,7 @@ public class ObjectsConverter implements Converter {
         return insn;
     }
 
+    // Calculates the struct's field size, i.e., including its inherited fields.
     private static int calculateFieldSize(ClassNode structClassNode, CompilationUnit compUnit) throws IOException {
         int fieldSize = structClassNode.fields.size();
         ClassNode currentClassNode = structClassNode;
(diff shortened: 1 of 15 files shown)


compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
```diff
@@ -321,48 +368,96 @@ private static boolean isCallToAnyStringBuilderMethod(AbstractInsnNode insn) {
...
 
-        NeoMethod calledNeoMethod;
-        String ctorMethodId = NeoMethod.getMethodId(ctorMethod, owner);
+        SuperNeoMethod calledNeoMethod;
+        String ctorMethodId = NeoMethod.getMethodId(ctorMethod, structClassNode);
+        int fieldSize = structClassNode.fields.size();
         if (compUnit.getNeoModule().hasMethod(ctorMethodId)) {
             // If the module already contains the converted ctor.
-            calledNeoMethod = compUnit.getNeoModule().getMethod(ctorMethodId);
+            calledNeoMethod = (SuperNeoMethod) compUnit.getNeoModule().getMethod(ctorMethodId);
```
csmuller, line 379: If the method was already compiled it is retrieved from the NeoModule. That's fine, but then the fieldSize is not set to the correct value if the class is inheriting fields from parent classes.

compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
```diff
@@ -321,48 +368,96 @@ private static boolean isCallToAnyStringBuilderMethod(AbstractInsnNode insn) {
                 && ((MethodInsnNode) insn).owner.equals(Type.getInternalName(StringBuilder.class));
     }
 
-    private static AbstractInsnNode convertConstructorCall(TypeInsnNode typeInsn,
-            MethodNode ctorMethod, ClassNode owner, NeoMethod callingNeoMethod,
-            CompilationUnit compUnit) throws IOException {
+    private static AbstractInsnNode convertStructConstructorCall(TypeInsnNode typeInsnNode, MethodNode ctorMethod,
```
csmuller, line 371: I'm pretty sure that parts of this can be refactored by using functionality from the `MethodConvertor`, i.e., passing the INVOKESPECIAL-MethodInsn to the `MethodConverter` and letting it do the work. Though, we can separate refactoring into another issue and move on with this one.

compiler/src/test-integration/java/io/neow3j/compiler/InstanceInheritanceIntegrationTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+import static io.neow3j.types.ContractParameter.array;
+import static io.neow3j.types.ContractParameter.hash160;
+import static io.neow3j.types.ContractParameter.integer;
+import static io.neow3j.types.ContractParameter.string;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+
+public class InstanceInheritanceIntegrationTest {
```
csmuller, line 29: Add a test that has a try-catch in the constructor.

csmuller (changes requested):

compiler/src/test-integration/java/io/neow3j/compiler/InstanceInheritanceIntegrationTest.java
mialbu: I'll add a test.

compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
csmuller: Will be handled in #741.

compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
```diff
@@ -124,6 +137,40 @@ public static void addStoreStaticField(FieldInsnNode fieldInsn, NeoMethod neoMet
...
+            }
+            // Now we're at the INVOKESPECIAL call and can convert the ctor method.
+            if (hasAnnotations(ctorMethod, Instruction.class, Instruction.Instructions.class)) {
+                processInstructionAnnotations(ctorMethod, callingNeoMethod);
+            }
+            return insn;
+        }
+    }
+
     public static AbstractInsnNode handleNew(AbstractInsnNode insn, NeoMethod callingNeoMethod,
```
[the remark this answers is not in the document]
csmuller: Will be handled in https://github.com/neow3j/neow3j/issues/741.

compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java
mialbu: You're absolutely right, I'll add an according test and fix this.

csmuller (approved):
🦾 good work!

PR #743, opened 2022-03-28
Updates for branching model change

mialbu:


RELEASE.md
```diff
@@ -1,30 +1,34 @@
...
-  - Set the name as "Release x.x.x"
-  - Set the correct milestone, project and a reviewer
-   
-- Review and merge the Release Pull Request, mainly checking for critical changes **or**
-changes that shouldn't be in the release
- 
-- Tag the main branch at the commit to be release (e.g., `3.2.1`)
- 
-- Right after merging, bump the version in the `build.gradle` file (`develop-3.x` branch)
+- Create a branch `release` from `main`.
```
gsmachado: Maybe, `release-<VERSION>`? or `release` is enough since this branch will be deleted after? Not strongly opinionated on this one.

RELEASE.md
```diff
@@ -1,30 +1,34 @@
...
+- Bump the neow3j version on that branch to the release version. Do a global search with the previous version number and
```
gsmachado: I generally dislike this.

I believe that the NEW version should be already on the development branch. Also because of SNAPSHOT releases.

So, I would bump the neow3j version RIGHT AFTER the release (directly on the `main` branch), and not BEFORE the release.

RELEASE.md
```diff
@@ -1,30 +1,34 @@
...
+- Create a Pull Request from `release` to `main` -- this is called a "Release Pull Request".
+    - Set the name as "Release x.x.x".
+    - Set the correct milestone, project and a reviewer.
+
+- Review the Release Pull Request, mainly checking for critical changes **or** changes that shouldn't be in the release.
+- Merge the Pull Request, and then, delete the `release` branch.
+
+- Tag the `main` branch at the commit to be release (e.g., `3.14.0`).
+
+- Right after merging, bump the version in the `build.gradle` and `Compiler.java` file on the `main` branch. Thus, also
```
gsmachado: Oh, here it goes. Why do we need to bump on earlier steps as well?!?

gsmachado (changes requested):

RELEASE.md
mialbu: `release` should be sufficient, since it's only used in a small timeframe anyways and then deleted.

RELEASE.md
mialbu: It's explained in right the next sentence that only the markdown files are meant in this step. I update the text to make it more clear.

Also, the bump in `build.gradle` and `Compiler.java` may be necessary, since we could have anticipate a non-breaking version number after the last release, but we may have breaking changes and require a different version number...

RELEASE.md
mialbu: See my comment above.

RELEASE.md
gsmachado: Sorry, my bad. I quickly read it and got confused.

gsmachado (approved):
In my point of view, everything looks fine!

Let's wait for @csmuller review. 💪

csmuller (approved):

PR #745, opened 2022-04-04
Minor changes the inheritance code

csmuller:


devpack/src/main/java/io/neow3j/devpack/annotations/Struct.java
```diff
@@ -4,7 +4,11 @@
 import java.lang.annotation.Target;
 
 /**
- * This annotation is used on class level to mark such as a struct.
+ * This annotation indicates that a class will be used as a NeoVM Struct.
+ * <p>
+ * Every Java class you intend to use in your smart contract class needs to be annotated with {code @Struct}
+ * in order for the compilation to be successful. This is because instances of Java classes are simple structs on the
+ * NeoVM. There is no notion of an object on NeoVM.
```
mialbu, line 11: I think this needs a bit rephrasing, since not *every* class needs that annotation, only the ones that are intended as NeoVM structs.

```suggestion
 * Every custom Java class you intend to use as a NeoVM Struct in your smart contract class must be annotated with
 * {code @Struct} in order for the compilation to be successful. This is because instances of Java classes are simple
 * structs on the NeoVM. There is no notion of an object on NeoVM.
```

mialbu (commented):
Great :+1:
I have only one comment/suggestion about the javadoc struct annotation.

devpack/src/main/java/io/neow3j/devpack/annotations/Struct.java
csmuller: Much better!

PR #749, opened 2022-04-08
Upgrade devpack-test fast-forward methods 

csmuller:


devpack-test/src/test-integration/java/io/neow3j/test/ModuleTest.java
```diff
@@ -112,4 +113,44 @@ public void transferTokensFromGenesisAccount() throws Throwable {
...
 
+    @Test
+    public void fastForward() throws Throwable {
+        BigInteger startCount = neow3j.getBlockCount().send().getBlockCount();
+        long startTime = neow3j.getBlock(startCount.subtract(BigInteger.ONE), false).send().getBlock().getTime();
+        ext.fastForward(60, 10);
+        BigInteger endCount = neow3j.getBlockCount().send().getBlockCount();
+        long endTime = neow3j.getBlock(endCount.subtract(BigInteger.ONE), false).send().getBlock().getTime();
+        assertThat(endCount, is(startCount.add(BigInteger.TEN)));
+        assertThat(endTime, is(greaterThanOrEqualTo(startTime + 60 * 1000))); // milliseconds
```
mialbu, line 124: Is there an imprecision? Shouldn't it be exactly 60 seconds, i.e., 'equals'? Same on lines 140 and 148.

mialbu (commented):
Nice feature! I've pushed some minor changes directly, and have one question commented to the tests.

devpack-test/src/test-integration/java/io/neow3j/test/ModuleTest.java
csmuller: Yes, I use `greaterThanOrEqualTo` because I expect imprecisions. Mainly because the underlying neo-express instance continues to produce blocks while the test is running. Also because I don't know if neo-express adds the  the time interval in `fastForward(...)` to the latest block or to the time at which the command was issued.

PR #752, opened 2022-04-12
Add support for NeoVM's ASSERT

### LINKED ISSUE #744: Support NeoVM's Assert Opcode
Currently, Java's `assert` statement is fully compiled to use NeoVM's `throw` opcode. However, developers should be enabled to use NeoVM's assert opcode which is cheaper than throwing an exception.

Suggestion: If Java's `assert` statements does not contain a message, it should be compiled to NeoVM's assert opcode.

mialbu:
Closes #744.

The use of Java's `assert` statement is refactored, so that it is now handled as the equivalent of the NeoVM's opcode `ASSERT`.

Further:
- Messages can no longer be added to the `assert` statement, since NeoVM's `ASSERT` does not consume a message. The compiler throws a `CompilerException` if a message exists.
- A hard coded `assert false` statement or throwing a `new AssertionError()` results in a `CompilerException`.

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
+++ b/compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
@@ -619,6 +619,9 @@ public class NeoMethod {
      * @return the last instruction.
      */
     public NeoInstruction getLastInstruction() {
+        if (this.instructions.size() == 0) {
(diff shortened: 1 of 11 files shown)


csmuller (approved):
Very good, it works!
I just added one more integration test.

PR #761, opened 2022-04-21
Fix handling of branch number in string switch-case compilation

csmuller:
The compiler failed compiling large switch-case constructs (more than 6 branches).


mialbu (approved):
:100:

PR #762, opened 2022-04-21
Fix bug when in compilation of Event without any arguments

csmuller:
Compilation for `Event` and `Event0Arg` failed.
I removed the latter because it was not intuitive to me when writing a contract.


gsmachado:
awesome we're finding these "little details that matter". 🎉

gsmachado (approved):
the famous "if statement that matters".

mialbu (approved):
Nice catch!

PR #769, opened 2022-05-04
Add option to allow transmission on fault

### LINKED ISSUE #767: Allow transmission of failing transactions in `TransactionBuilder`
The `TransactionBuilder` always throws an exception when building a transaction that holds a script leading to a VM FAULT. This is due to the lines at https://github.com/neow3j/neow3j/blob/af237d1ac9d93b6e1611aa749a5e781e6280d1ac/core/src/main/java/io/neow3j/transaction/TransactionBuilder.java#L384-L387

Sometimes we want a transaction to be invoked even if neow3j knows that it will fail, e.g., for contract tests. Therefore, we need a method to disable the above exception. I propose something like the following:

```java
public void setTransmitOnVMFault() { 
    this.transmitOnFault = true;
}
```

And then an extra check before throwing the exception: 
```java
if (response.getResult().hasStateFault() && !transmitOnFault) {
```

This can be implemented as a non-breaking change.

mialbu:
Closes #767

### DIFF
--- a/core/src/main/java/io/neow3j/protocol/Neow3j.java
+++ b/core/src/main/java/io/neow3j/protocol/Neow3j.java
@@ -48,6 +48,30 @@ public abstract class Neow3j implements Neo, Neow3jRx {
     public abstract void shutdown();
 
     /**
+     * @return true if transmission is allowed when the provided script leads to a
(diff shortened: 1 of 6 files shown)


csmuller (approved):
💯

PR #774, opened 2022-05-09
Refactor tests with assertThrows

### LINKED ISSUE #763: Incorrect use of assertThrows
The method `Assert.assertThrows` can take a string value to provide a message if the assertion fails. However, in some of the tests this message was assumed to be the message of the expected thrown exception.

E.g., in `SignerTest.failAddingRuleToGlobalSigner()` only the exception class is checked:

```java
assertThrows("Trying to set witness rules on a Signer with global scope.",
        SignerConfigurationException.class,
        () -> AccountSigner.global(acc).setRules(rule));
```

In order to assert the exception message, the above code can be changed to the following:

```java
SignerConfigurationException thrown = assertThrows(SignerConfigurationException.class,
        () -> AccountSigner.global(acc).setRules(rule));
assertThat(thrown.getMessage(), is("Trying to set witness rules on a Signer with global scope."));
```

Every use of `assertThrown` should be checked in neow3j to ensure the tests' correctness.

mialbu:
Resolves #763

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/CompilerException.java
+++ b/compiler/src/main/java/io/neow3j/compiler/CompilerException.java
@@ -2,6 +2,8 @@ package io.neow3j.compiler;
 
 import org.objectweb.asm.tree.ClassNode;
 
+import static java.lang.String.format;
+
 public class CompilerException extends RuntimeException {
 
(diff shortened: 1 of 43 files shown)


csmuller (approved):
jiffy

PR #779, opened 2022-05-17
Enforce String type for String concatenation

### LINKED ISSUE #760: Fail compilation when developer tries to concatenate string with other type
An example of this problem is the following:
Developers often want to concatenate strings with integers to print an event, log, or exception message. But, the neow3j compiler does not support string concatenation with other types as Java does.

So either we add extra functionality such that the compiler supports such concatenations or we make it clear at compile-time that it is not supported by throwing a compiler exception.

mialbu:
Closes #760

### DIFF
--- a/compiler/src/test-integration/java/io/neow3j/compiler/StringConcatenationTest.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/StringConcatenationTest.java
@@ -51,15 +51,39 @@ public class StringConcatenationTest {
     }
 
     @Test
+    public void concatInStaticVariableWithChar() throws IOException {
+        NeoInvokeFunction response = ct.callInvokeFunction(testName);
+        assertThat(response.getInvocationResult().getStack().get(0).getString(), is("onetwon"));
+    }
+
+    @Test
     public void concatStringWithMethodReturn() throws IOException {
         InvocationResult response = ct.callInvokeFunction(testName).getInvocationResult();
         assertThat(response.getStack().get(0).getString(), is("awesome token"));
     }
 
     @Test
+    public void concatStringWithMethodReturnChar() throws IOException {
+        InvocationResult response = ct.callInvokeFunction(testName).getInvocationResult();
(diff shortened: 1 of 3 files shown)


compiler/src/test-integration/java/io/neow3j/compiler/StringConcatenationTest.java
```diff
@@ -64,6 +98,31 @@ public static String concatInStaticVariable() {
             return staticString;
         }
 
+        public static String concatStringWithMethodReturn() {
```
gsmachado, line 101: does it worth adding tests with all Java built-in raw types?

Like: `int` (we already have test for this), `byte`, `short`, `long`, `float`, `double`, `boolean`, and `char`? 

Also, maybe, array of these types?

wdyt?

gsmachado (commented):
Before I approve: just a question about tests. 😄

compiler/src/test-integration/java/io/neow3j/compiler/StringConcatenationTest.java
mialbu: @gsmachado 
- Trying to concatenate arrays (except `byte[]`) results in an exception on the NeoVM (`Specified cast is not valid`). Therefore, no array types can be used in string concatenations in neow3j. I have added a UT for this in the class `StringLiteralHelperTest.java`.

- `byte`, `byte[]`, `long`, `short` all have the same behaviour as `int`. There are UTs in place for `int` concatenation with strings that are expected to throw an exception. Hence, there's no need to have more tests for these types atm.

- `char` works the same way as `String`. I've added integration tests for the primitive type as well as for the wrapper class (`java/lang/Character`).

- Last but not least: Neow3j does not support float and double in the first place. So, no need for further tests there. The compilation would fail as soon as it notices such types present anyways.

gsmachado: great! All fine, then.

gsmachado (approved):
Everything is clear now. 🎉  Ready to go. 💪

PR #794, opened 2022-06-07
Update for Neo v3.3.0

### LINKED ISSUE #786: Update for Neo v3.3.0
- [x] #759
- [x] #787
- [x] #788
- [x] #789
- [x] #790
- [x] #791
- [x] #792

mialbu:
Closes #786

### DIFF
--- a/compiler/src/test-integration/java/io/neow3j/compiler/ContractTestRule.java
+++ b/compiler/src/test-integration/java/io/neow3j/compiler/ContractTestRule.java
@@ -18,6 +18,9 @@ import io.neow3j.transaction.Signer;
 import io.neow3j.transaction.Transaction;
 import io.neow3j.transaction.TransactionBuilder;
 import io.neow3j.transaction.Witness;
+import io.neow3j.transaction.witnessrule.CalledByEntryCondition;
+import io.neow3j.transaction.witnessrule.WitnessAction;
+import io.neow3j.transaction.witnessrule.WitnessRule;
 import io.neow3j.types.ContractParameter;
 import io.neow3j.types.Hash160;
 import io.neow3j.types.Hash256;
@@ -37,6 +40,7 @@ import java.util.List;
 
 import static io.neow3j.crypto.Sign.signMessage;
 import static io.neow3j.test.TestProperties.client1AccountWIF;
+import static io.neow3j.test.TestProperties.client2AccountWIF;
 import static io.neow3j.test.TestProperties.defaultAccountWIF;
 import static io.neow3j.transaction.Witness.createMultiSigWitness;
 import static io.neow3j.utils.ArrayUtils.reverseArray;
@@ -57,6 +61,7 @@ public class ContractTestRule implements TestRule {
     private Account defaultAccount;
     private Account committee;
     private Account client1;
+    private Account client2;
     private Neow3j neow3j;
     private SmartContract contract;
     private Hash256 deployTxHash;
@@ -98,6 +103,7 @@ public class ContractTestRule implements TestRule {
         committee = Account.createMultiSigAccount(
                 asList(defaultAccount.getECKeyPair().getPublicKey()), 1);
         client1 = Account.fromWIF(client1AccountWIF());
+        client2 = Account.fromWIF(client2AccountWIF());
         neow3j = Neow3j.build(new HttpService(containerURL, true));
         waitUntilBlockCountIsGreaterThanZero(neow3j);
     }
@@ -150,6 +156,10 @@ public class ContractTestRule implements TestRule {
 
     public Account getClient1() {
         return client1;
+    }
+
+    public Account getClient2() {
+        return client2;
     }
 
     public Neow3j getNeow3j() {
@@ -314,6 +324,10 @@ public class ContractTestRule implements TestRule {
         TransactionBuilder b = contract.invokeFunction(function, params.toArray(new ContractParameter[0]));
         List<Signer> signerList = new ArrayList<>(asList(additionalSigners));
         if (signAsCommittee) {
+//            signerList.add(AccountSigner.none(committee).setAllowedContracts(getContract().getScriptHash()));
+            WitnessRule rule = new WitnessRule(WitnessAction.ALLOW, new CalledByEntryCondition());
+//            AccountSigner.none(committee).setRules(rule);
+//            signerList.add(AccountSigner.none(committee).setRules(rule));
             signerList.add(AccountSigner.global(committee));
             Signer[] modifiedSigners = signerList.toArray(new Signer[0]);
             tx = b.signers(modifiedSigners).firstSigner(committee).getUnsignedTransaction();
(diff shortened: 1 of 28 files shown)


contract/src/main/java/io/neow3j/contract/NeoToken.java
```diff
@@ -185,7 +188,7 @@ public List<ECPublicKey> getCommittee() throws IOException {
      * <p>
      * The vote count is based on the summed up NEO balances of the respective candidate's voters.
      *
-     * @return the candidate public keys and their corresponding vote count.
+     * @return the validator public keys and their corresponding vote count.
```
csmuller, line 191: I don't agree with the word change here. The list returned by the neo-node contains candidates and the current committee members. They are not necessarily validators.
Additionally, in the "worst case" this returns 256 candidates.

contract/src/main/java/io/neow3j/contract/NeoToken.java
```diff
@@ -197,30 +200,72 @@ public Map<ECPublicKey, BigInteger> getCandidates() throws IOException {
...
+        if (!candItem.getType().equals(STRUCT)) {
+            throw new UnexpectedReturnTypeException(candItem.getType(), STRUCT);
+        }
+        ECPublicKey key = extractPublicKey(candItem.getList().get(0));
+        StackItem nrItem = candItem.getList().get(1);
+        if (!nrItem.getType().equals(INTEGER)) {
+            throw new UnexpectedReturnTypeException(nrItem.getType(), INTEGER);
+        }
+        candidates.put(key, nrItem.getInteger());
+    }
```
csmuller, line 256: Minor detail: The candidates map should be created inside of this method instead of taking it as a parameter and modifying it. It's not bad in this case but usually I wouldn't want a method to have suche "side effects", i.e., taking a value as parameter and modifying it without the caller expecting it.

devpack/src/main/java/io/neow3j/devpack/contracts/NeoToken.java
```diff
@@ -48,10 +49,20 @@ public class NeoToken extends FungibleToken {
     public static native boolean vote(Hash160 scriptHash, ECPoint candidatePubKey);
 
     /**
-     * @return the list of registered candidates.
+     * @return the list of the current validators.
```
csmuller, line 52: More precisely this returns the first 256 registered candidates. Again I don't agree with "validators" here.

devpack/src/main/java/io/neow3j/devpack/Signer.java
```diff
@@ -0,0 +1,71 @@
...
+     *
+     * @param signer other signer to compare to.
+     * @return true if all fields of the two contracts are equal. False otherwise.
+     */
+    public boolean equals(Signer signer) {
+        if (this == signer) {
+            return true;
+        }
+        return account == signer.account && allowedContracts == signer.allowedContracts &&
+                allowedGroups == signer.allowedGroups;
```
csmuller, line 68: This will check for equality by reference for the fields `allowedContracts`, `allowedGroups` and `witnessRules` (which is missing here). That is not what the user would expect. If you want to support `equals` here, you will have to iterate through the arrays. This also requires implementing an `equals` method for `WitnessRule` and `WitnessRuleCondition`.

Check the `equals` method in `io.neow3j.devpack.Manifest`. It gives an example of how tedious a correct equals method can be.

devpack/src/main/java/io/neow3j/devpack/Signer.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * The groups that are allowed by the witness. Only available when the corresponding signer has a witness scope
+     * {@link io.neow3j.transaction.WitnessScope#CUSTOM_GROUPS}.
+     */
+    public ECPoint[] allowedGroups;
+
+    /**
+     * The groups that are allowed by the witness. Only available when the corresponding signer has a witness scope
+     * {@link io.neow3j.transaction.WitnessScope#CUSTOM_GROUPS}.
```
csmuller, line 37: Witness scope should be `WITNESS_RULES` not `CUSTOM_GROUPS`

core/src/main/java/io/neow3j/script/InteropService.java
```diff
@@ -17,8 +17,8 @@ public enum InteropService {
     SYSTEM_CONTRACT_CALL("System.Contract.Call", 1 << 15),
     SYSTEM_CONTRACT_CALLNATIVE("System.Contract.CallNative", 0),
     SYSTEM_CONTRACT_GETCALLFLAGS("System.Contract.GetCallFlags", 1 << 10),
-    SYSTEM_CONTRACT_CREATESTANDARDACCOUNT("System.Contract.CreateStandardAccount", 1 << 8),
-    SYSTEM_CONTRACT_CREATEMULTISIGACCOUNT("System.Contract.CreateMultisigAccount", 1 << 8),
+    SYSTEM_CONTRACT_CREATESTANDARDACCOUNT("System.Contract.CreateStandardAccount", 0),
```
csmuller, line 20: Screenshot 2022-06-21 at 09 38 47
Based on the changes in the neo code, we can actually set the value to 1 << 15. See `CreateStandardAccount` in `Application.Contract.cs`. I was mistaken, it is still fixed after the hardfork. The value is according to the `SYSTEM_CRYPTO_CHECKSIG`. So we could use
```java
SYSTEM_CONTRACT_CREATESTANDARDACCOUNT("System.Contract.CreateStandardAccount", SYSTEM_CRYPTO_CHECKSIG.price)
```

csmuller (changes requested):

contract/src/main/java/io/neow3j/contract/NeoToken.java
mialbu: You're totally right. I'll change it back.

contract/src/main/java/io/neow3j/contract/NeoToken.java
mialbu: Noted, I'll improve that part.

devpack/src/main/java/io/neow3j/devpack/contracts/NeoToken.java
mialbu: You're right, I'll change it.

compiler/src/test-integration/java/io/neow3j/compiler/ContractTestRule.java
```diff
@@ -314,6 +324,10 @@ public Hash256 invokeFunction(String function, List<ContractParameter> params, S
         TransactionBuilder b = contract.invokeFunction(function, params.toArray(new ContractParameter[0]));
         List<Signer> signerList = new ArrayList<>(asList(additionalSigners));
         if (signAsCommittee) {
+//            signerList.add(AccountSigner.none(committee).setAllowedContracts(getContract().getScriptHash()));
```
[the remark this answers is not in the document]
mialbu: I committed that by mistake, I'll remove it.

devpack/src/main/java/io/neow3j/devpack/Signer.java
mialbu: I'll change that.

devpack/src/main/java/io/neow3j/devpack/Signer.java
mialbu: I've added an improved implementation of it, have a look at the commit 1f5cc85.

mialbu:
All tests pass except integration tests for `findStates` RPC. This is due to an issue on the neo-modules and will be fixed in the PR https://github.com/neo-project/neo-modules/pull/734.

PR #809, opened 2022-06-24
Restructure integration tests

### LINKED ISSUE #689: Refactor integration tests in neow3j

IMHO, it would be nice if we could refactor the integration tests in all modules.

Some reasons are:
- at the moment we have a bunch of different ways of how we perform integration tests in neow3j
- some integration tests might depend on the `contract` module, which sometimes is not available due to redundant dependencies

Maybe it would be nice to put the integration tests in a layer above all neow3j modules, that we can use anything we have in neow3j.

mialbu:
Closes #689


mialbu:
`Noew3jExpressIntegrationTest` is failing. @gsmachado and I haven't found the problem. Could you take a look, @csmuller?

Also, the tests for the `findstates` RPC are still failing, but pass with Neo `v3.3.1` (tested locally).

csmuller (approved):
Fails only because of the state-related integration tests. Ready to merge.

PR #813, opened 2022-06-27
Add CallFlags annotation to devpack

### LINKED ISSUE #778: Add `@CallFlags` annotation
When calling a contract method using a wrapper class, the call flag `ALL` is automatically added by default. That call flag allows the called contract every action (i.e., fire notifications, calling another contract, writing to storage, etc.). There should be an option to set the call flags when using contract wrapper classes.

Neow3j could provide an option to mark contract calls with the desired call flags when using a contract wrapper class. This could be achieved with the use of a new annotation `@CallFlags()`.

- Compiler

mialbu:
Closes #778


csmuller (approved):
💪

PR #817, opened 2022-06-30
Update NFT integration tests

mialbu:
With the recent changes, the NFT integration tests no longer need to statically set the deployed contract hash. With the changes in this PR, the contract hashes are now set dynamically and after a version bump, these tests no longer need separate updating.

Accordingly, I have updated the `RELEASE.md` which no longer needs this step.

Finally, I've removed the `abstract` modifier in the classes `ContractInterface`, `NonFungibleToken`, and `Token`. That way methods held by these classes can also be used without using/creating a specific subtype of them (e.g., `NonDivisibleNonFungibleToken` if it doesn't matter whether the called NFT is divisible or not...).


csmuller (approved):
jiffy

PR #826, opened 2022-07-07
Support Hash and ECPoint from Constant String

### LINKED ISSUE #801: Better way to create Hash160 from script hash string
Currently the dev has to do the following to create a `Hash160` from a big-endian script hash string.
```java
new Hash160(Helper.reverse(StringLiteralHelper.hexToBytes("0xfffdc93764dbaddd97c48f252a53ea4643faa3fd").toByteArray()))
```
Crazy, right?

It would be preferable to have the following.
```java
new Hash160("0xfffdc93764dbaddd97c48f252a53ea4643faa3fd")
```

The same should be supported for `Hash256` and `ECPoint`.

mialbu:
Closes #801

### DIFF
--- a/compiler/src/test/java/io/neow3j/compiler/HashTest.java
+++ b/compiler/src/test/java/io/neow3j/compiler/HashTest.java
@@ -1,19 +1,23 @@
 package io.neow3j.compiler;
 
-import static io.neow3j.devpack.StringLiteralHelper.hexToBytes;
-import static java.util.Arrays.asList;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
-
 import io.neow3j.compiler.sourcelookup.MockSourceContainer;
 import io.neow3j.devpack.Hash160;
 import io.neow3j.devpack.Hash256;
+import io.neow3j.script.OpCode;
 import io.neow3j.types.ContractParameterType;
+import org.junit.Test;
 
 import java.io.File;
 import java.io.IOException;
 
-import org.junit.Test;
+import static io.neow3j.devpack.StringLiteralHelper.hexToBytes;
+import static io.neow3j.utils.Numeric.hexStringToByteArray;
+import static io.neow3j.utils.Numeric.reverseHexString;
+import static java.util.Arrays.asList;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertThrows;
 
 public class HashTest {
 
@@ -21,7 +25,7 @@ public class HashTest {
     public void hash160ReturnTypeShouldAppearAsHash160InManifest() throws IOException {
         CompilationUnit compUnit = new Compiler().compile(HashTestContract.class.getName());
         assertThat(compUnit.getManifest().getAbi().getMethods().get(0).getName(),
-                is("methodReturningHash160"));
+                is("hash160FromBytes"));
         assertThat(compUnit.getManifest().getAbi().getMethods().get(0).getReturnType(),
                 is(ContractParameterType.HASH160));
     }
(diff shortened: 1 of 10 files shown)


int-tests/src/test-integration/java/io/neow3j/compiler/HashIntegrationTest.java
```diff
@@ -257,6 +271,14 @@ public static Hash256 hash256FromStringLiteral() {
             return new Hash256(
                     StringLiteralHelper.hexToBytes("03b4af8d061b6b320cce6c63bc4ec7894dce107b000000000000000000000000"));
         }
+
+        public static io.neow3j.devpack.Hash160 hash160FromString() {
+            return new io.neow3j.devpack.Hash160("0xcc5e4edd9f5f8dba8bb65734541df7a1c081c67b");
```
csmuller, line 276: Do we also support strings without the "0x" prefix?

int-tests/src/test-integration/java/io/neow3j/compiler/ECPointIntegrationTest.java
```diff
@@ -123,6 +138,15 @@ public static ByteString ecPointAsByteString(ECPoint ecPoint) {
             return ecPoint.toByteString();
         }
 
+        public static ECPoint ecPointFromString() {
+            return new ECPoint("0x03b4af8d061b6b320cce6c63bc4ec7894dce107bfc5f5ef5c68a93b4ad1e136816");
```
csmuller, line 142: Do we also support strings without the "0x" prefix?

compiler/src/test/java/io/neow3j/compiler/HashTest.java
```diff
@@ -54,20 +58,92 @@ public void hash256ReturnTypeShouldAppearAsHash256InDebugInfo() throws IOExcepti
...
+
+    @Test
+    public void testInvalidHash256FromString() {
+        IllegalArgumentException thrown = assertThrows(IllegalArgumentException.class,
+                () -> new Compiler().compile(InvalidHash256.class.getName()));
+        assertThat(thrown.getMessage(), is("Hash must be 32 bytes long but was 27 bytes."));
+    }
+
+    @Test
+    public void testHashFromStringWithNonConstantString() {
```
csmuller, line 102: There should be a test just like this for ECPoint in `ECPointTest`.

csmuller (changes requested):

int-tests/src/test-integration/java/io/neow3j/compiler/HashIntegrationTest.java
mialbu: Yes, a `types.Hash160` is initialised when compiling that string. So, the prefix does not matter. See here: https://github.com/neow3j/neow3j/blob/f45ce15d23cdfc6820a95c97b54d601147aa06d7/compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java#L334-L339

int-tests/src/test-integration/java/io/neow3j/compiler/ECPointIntegrationTest.java
mialbu: Yes, a `types.ECPublicKey` is initialised when compiling that string. So, the prefix does not matter. See here: https://github.com/neow3j/neow3j/blob/f45ce15d23cdfc6820a95c97b54d601147aa06d7/compiler/src/main/java/io/neow3j/compiler/converters/ObjectsConverter.java#L348-L353

compiler/src/test/java/io/neow3j/compiler/HashTest.java
mialbu: This basically tests the method `ObjectsConverter.checkForConstantStringArgument()` which is called when handling `Hash160`, `Hash256`, or `ECPoint` instances with a string argument. So, do you think I should add a test for all three anyways?

csmuller: If we were testing the `checkForConstantStringArgument` directly one test would be enough. But here we're testing on a higher level. I think that it's necessary to test it for all three classes because we might change the behavior of specifically one in the future. Then a specific test will tell us what the expected behavior is/was.

mialbu: Good point, I'll add tests for all three.

csmuller (approved):
👏

PR #827, opened 2022-07-07
Allow uncompressed public key in ECPublicKey constructor

### LINKED ISSUE #803: One should also be able to instantiate ECPublicKey with an uncompressed key
the `ECPublicKey(byte[] publicKey)` constructor should also allow uncompressed public keys. This is currently not possible because of this check: `if (publicKey.length != NeoConstants.PUBLIC_KEY_SIZE)`.
We can simply remove the check, because bouncy castle will throw an exception if the keys length is incorrect.

mialbu:
Closes #803


csmuller (dismissed):
👏

csmuller (approved):

PR #838, opened 2022-07-22
Fix String Concatenation

### LINKED ISSUE #781: Handle Javac Optimisation of String Concatenations
**Description of the bug**
It appears that when I am trying to concatenate two Strings, when one is from the storage, then the compiler complains with the following message:

```
Inherited methods that are not specifically implemented are not supported. Implement the method 'toString' without a 'super' call to the class Object to use it.
io.neow3j.compiler.CompilerException: Inherited methods that are not specifically implemented are not supported. Implement the method 'toString' without a 'super' call to the class Object to use it.
        at io.neow3j.compiler.Compiler.throwIfObjectIsOwner(Compiler.java:603)
        at io.neow3j.compiler.Compiler.handleInsn(Compiler.java:584)
        at io.neow3j.compiler.NeoMethod.convert(NeoMethod.java:481)
        at io.neow3j.compiler.converters.MethodsConverter.handleUncachedMethodCall(MethodsConverter.java:224)
```

**To Reproduce**
1. create new smart contract
1.  try to concatenate two strings: map.getString(key) + "foo";
2. or: map.get(key).toString() + "foo"
3. compile 

**Expected behavior**
It should be possible to concatenate two strings like this without compiling issues.

**Platform and Software**
 - neow3j 3.17

mialbu:
Closes #781


csmuller (approved):
LGTM... ;) actually nicely done with the precompiled class files in the tests.

mialbu:
> LGTM... ;) actually nicely done with the precompiled class files in the tests.

Thanks :)

gsmachado:
@mialbu AWESOME pull request. 🎉 

Great set of tests as well, with the pre-compiled `.class` files. 💪

gsmachado:
@mialbu should we delete the branch now that is merged? 🤔

PR #839, opened 2022-07-22
Add valid hex check in Numeric class

### LINKED ISSUE #834: Numeric.hexStringToByteArray(...) should check for hex string validity
`Numeric.hexStringToByteArray(...)` does no check if the argument is a valid hex string.
I.e., it is possible to do the following without an exception:
```java
Numeric.hexStringToByteArray("degen apes")
```

The method should check if the argument is a valid hex string, like this:
```java
if (!isValidHexString(input)) {
    throw new IllegalArgumentException("Invalid hex string.");
}
```
Additionally, a test for this should be added to `NumericTest`.

mialbu:
Closes #834

### DIFF
--- a/core/src/main/java/io/neow3j/utils/Numeric.java
+++ b/core/src/main/java/io/neow3j/utils/Numeric.java
@@ -229,6 +229,7 @@ public final class Numeric {
     }
 
     public static byte[] hexStringToByteArray(String input) {
+        checkAndThrowIsValidHexString(input);
         String cleanInput = cleanHexPrefix(input);
 
         int len = cleanInput.length();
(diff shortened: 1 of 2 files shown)


gsmachado (dismissed):
small but important change. 👍 💪 💯

gsmachado:
does anyone know why CI-CD tests are failing? 🤔

csmuller:
The test that is failing is testing with a hex string that is now considered invalid. And I think it's correct like this. Only hex strings with an even number of characters should be accepted by `hexStringToByteArray`. @mialbu, please change the test (or add a new one that checks if an exception is thrown for one or two cases of invalid hex strings.

csmuller (approved):

PR #840, opened 2022-07-25
Use NEWARRAY0 for empty arrays

### LINKED ISSUE #799: Compiler optimization: Use the opcode `NEWARRAY0` for empty arrays
In some cases when an empty array is created, the neow3j compiler first pushes the value `0` and then uses the opcode `NEWARRAY`. This should be optimized by using the opcode `NEWARRAY0`.

mialbu:
Closes #799


csmuller (approved):
👍

csmuller (approved):

PR #851, opened 2022-08-03
Upgrade dependencies

### LINKED ISSUE #845: Upgrade dependencies

mialbu:
Closes #845


core/src/main/java/io/neow3j/protocol/http/HttpService.java
```diff
@@ -255,15 +255,14 @@ private static OkHttpClient createOkHttpClient() {
...
+            logging.level(HttpLoggingInterceptor.Level.BODY);
             builder.addInterceptor(logging);
         }
     }
 
     @Override
     protected InputStream performIO(String request) throws IOException {
-
-        RequestBody requestBody = RequestBody.create(JSON_MEDIA_TYPE, request);
+        RequestBody requestBody = RequestBody.create(request, JSON_MEDIA_TYPE);
```
gsmachado, line 265: Seriously? They changed the order of params keeping the same method name?! 🤯

gsmachado (dismissed):
It looks fine to me... but still don't believe they changed the order of the parameters and kept the same method name. This is almost a crime. It's not an actual crime because this change was done in a major version upgrade... 👍

core/src/main/java/io/neow3j/protocol/http/HttpService.java
mialbu: Yes, that's actually the case :laughing:

gsmachado (approved):
Alright... **Assuming** the integration tests passed (locally or remotely), as well as the unit tests, then it's good to go.

PR #852, opened 2022-08-04
Add method to unwrap iterator

### LINKED ISSUE #842: Provide method to get iterator values without sessions enabled
If public nodes don't have sessions enabled, retrieving values of an iterator becomes cumbersome.

Neow3j should provide a workaround that gets the iterator values on vm-level and return an array filled with these values like the method `CreateCallAndUnwrapIteratorScript` in neo-go.

mialbu:
Closes #842

### DIFF
--- a/contract/src/main/java/io/neow3j/contract/SmartContract.java
+++ b/contract/src/main/java/io/neow3j/contract/SmartContract.java
@@ -200,6 +200,9 @@ public class SmartContract {
      * {@link InteropInterfaceStackItem} as a return type.
      * <p>
      * Traverse the returned iterator with {@link Iterator#traverse(int)} to retrieve the iterator items.
+     * <p>
+     * In order to traverse the returned iterator, sessions need to be enabled on the Neo node. If sessions are
+     * disabled on the Neo node, use {@link SmartContract#callFunctionAndUnwrapIterator(String, List, int, Signer...)}.
      *
      * @param function the function to call.
      * @param params   the contract parameters to include in the call.
@@ -217,6 +220,9 @@ public class SmartContract {
      * <p>
      * Traverse the returned iterator with {@link Iterator#traverse(int)} to retrieve the iterator items with
      * provided mapper applied to each item.
+     * <p>
+     * In order to traverse the returned iterator, sessions need to be enabled on the Neo node. If sessions are
+     * disabled on the Neo node, use {@link SmartContract#callFunctionAndUnwrapIterator(String, List, int, Signer...)}.
      *
      * @param mapper   the function to apply on the stack items in the iterator.
      * @param function the function to call.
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/neow3j/script/ScriptBuilder.java
```diff
@@ -422,4 +424,96 @@ public static byte[] buildContractHashScript(Hash160 sender, long nefCheckSum, S
...
+     *
+     * @param contractHash           the script hash of the contract to call.
+     * @param method                 the method to call.
+     * @param params                 the parameters that will be used in the call. Need to be in correct order.
+     * @param callFlags              the call flags to use for the contract call.
+     * @param maxIteratorResultItems the maximal number of iterator result items to include in the array. This value
+     *                               must not exceed NeoVM limits.
+     * @return the script.
+     */
+    public static byte[] buildContractCallAndUnwrapIterator(Hash160 contractHash, String method,
```
csmuller, line 465: Very nice in-code comments! 💚

int-tests/src/test-integration/java/io/neow3j/compiler/DivisibleNFTIntegrationTest.java
```diff
@@ -62,6 +64,21 @@ public void testOwnerOf() throws IOException {
         assertThat(iter.get(1).getAddress(), is("NhxK1PEmijLVD6D4WSuPoUYJVk855L21ru"));
     }
 
+    @Test
+    public void testOwnerOfWithoutSessions() throws IOException {
```
csmuller, line 68: What happens when you try to get more items than available in the iterator?

csmuller (changes requested):

int-tests/src/test-integration/java/io/neow3j/compiler/DivisibleNFTIntegrationTest.java
mialbu: All items are returned. The cycle in the VM script is exited once the syscall `IteratorNext` returns false.

I'll add a test where the `maxIteratorReturnValues` is greater than the items in the iterator.

contract/src/main/java/io/neow3j/contract/SmartContract.java
```diff
@@ -282,6 +305,35 @@ public List<StackItem> callFunctionAndTraverseIterator(String function, Contract
...
+     * Sends an {@code invokescript} RPC call. The script that is built calls a contract function that is expected to
+     * return an iterator. The script then traverses this iterator and adds the read items to an array which is
+     * returned once the iterator is completely traversed or the array reaches the size of
+     * {@code maxIteratorResultItems}.
+     * <p>
+     * Use this to retrieve iterator values in interaction with an RPC server that has sessions disabled.
+     *
+     * @param function               the function to call.
+     * @param params                 the contract parameters to include in the call.
+     * @param maxIteratorResultItems the maximal number of iterator result items to include in the array. This value
```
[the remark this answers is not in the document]
mialbu: It is limited by `MaxStackItems` and `MaxGasInvoke`, so there's no specific default value that we could use here.

As from our disscussion, I will leave the current implementation as is.

csmuller (approved):
👍

PR #859, opened 2022-08-10
Migrate tests to JUnit 5

### LINKED ISSUE #822: Migrate integration tests to JUnit 5
The new int-test module uses JUnit 4 as its test framework, but the `ModuleTest` test class is based on JUnit 5. Therefore, it is not executed when running the tests of the int-test module.

One solution would be to separate the tests in int-test into different source stets and configure them separately under different `testIntegration` Gradle tasks. Then we'd have to collect them in one main `testIntegration` task that dependsOn all of them.

The other solution is to update all integration tests to JUnit 5. That's what this issue wants.

mialbu:
Closes #822 

With these changes all tests run with JUnit 5 except for the `gradle-plugin` integration tests.


core/src/test/java/io/neow3j/crypto/MnemonicUtilsTest.java
```diff
@@ -35,66 +32,51 @@ public class MnemonicUtilsTest {
...
-        this.initialEntropy = Hex.decode(initialEntropy);
-        this.mnemonic = mnemonic;
-        this.seed = Hex.decode(seed);
-    }
-
-    @Test
-    public void generateMnemonicShouldGenerateExpectedMnemonicWords() {
-        String actualMnemonic = MnemonicUtils.generateMnemonic(initialEntropy);
+    @ParameterizedTest
+    @MethodSource("data")
```
csmuller, line 59: image

core/src/test/java/io/neow3j/protocol/RequestTester.java
```diff
@@ -9,21 +9,23 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.TestInstance;
 
 import static io.neow3j.protocol.http.HttpService.JSON_MEDIA_TYPE;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.Assert.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
csmuller, line 20: image

int-tests/src/test-integration/java/io/neow3j/compiler/AccountIntegrationTest.java
```diff
@@ -18,26 +21,30 @@
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 
+@Testcontainers
```
csmuller, line 24: The `@Testcontainers` annotation is obsolete because the container life cycle is managed by the `ContractTestExtension`. This might not be the most elegant way but it's how we implemented it and it's easier to keep it that way. Please remove the annotation on all similar test classes - i.e., everywhere the `ContractTestExtension` is used but not a container as a test class member - and also from `io.neow3j.compiler.ContractTestExtension` itself.

Instead of the extension we could extend a `TestContainer` class and try to pack the `ContractTestExtension` functionality in there. Then the `@Testcontainers` annotation would handle the life cycle. But, I don't think that's necessary at the moment.

int-tests/src/test-integration/java/io/neow3j/compiler/ByteArrayTest.java
```diff
@@ -16,14 +17,18 @@
...
+    private String testName;
 
-    @ClassRule
-    public static ContractTestRule ct = new ContractTestRule(
-            ByteArrayTestContract.class.getName());
+    @RegisterExtension
+    public static ContractTestExtension ct = new ContractTestExtension(ByteArrayTestContract.class.getName());
+
+    @BeforeEach
+    void init(TestInfo testInfo) {
```
csmuller, line 29: image

int-tests/src/test-integration/java/io/neow3j/protocol/TransactionBuilderIntegrationTest.java
```diff
@@ -38,10 +39,11 @@
...
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
+@Testcontainers
```
csmuller, line 46: I think this is the correct case to use this annotation since you have a test container as a member of the class.

csmuller (changes requested):

int-tests/src/test-integration/java/io/neow3j/compiler/AccountIntegrationTest.java
mialbu: True, I missed that fact. I'll remove the annotations accordingly.

csmuller (dismissed):

csmuller:
Waiting with the merge till the Neo 3.4.0 PR is merged.

PR #889, opened 2022-11-01
Fix ContractParameter equals method

### LINKED ISSUE #887: ContractParameters: bug in `equals` method
The `equals` method of `ContractParameter` tries to cast the value of Hash160/Hash256 parameters to `byte[]`, but, the value is of type Hash160/Hash256.

This wouldn't happen if we did #833 😏

csmuller:
I had to fix this now, because it after all got more problematic for the GrantShares code.

Closes #887


gsmachado:
@csmuller wow, well done. I ❤️ the test.

csmuller:
> @csmuller wow, well done. I ❤️ the test.

haha, yes, but the implementation is actually erroneous  😅 Fixing it now.

mialbu (approved):
Nice work! 💯

PR #915, opened 2023-01-23
Add NeoFS support for armhf/armv7

gsmachado:


mialbu (approved):
Even though support for NeoFS is currently stalled in neow3j, I think we can merge this and close this PR. Do you agree, @gsmachado?

gsmachado:
> Even though support for NeoFS is currently stalled in neow3j, I think we can merge this and close this PR. Do you agree, @gsmachado?

Yes, agree.

PR #930, opened 2023-04-18
Utility methods for accessing items by index directly

### LINKED ISSUE #924: Add utility methods to get first items
Neow3j has many methods to return a list of objects where the first item is important. This results in many lines of code that contain `.get(0)`.

In order to simplify this, neow3j could add two kind of methods:
- A method that accesses the first index, e.g., `.getFirstStackItem()`
- A method that takes an index as a parameter to directly retrieve the item without having to call `.get(i)`. This would make the code appear more succinct. For example: `.getStackItem(2)`, instead of `getStack().get(2)`

- SDK

mialbu:
Resolves #924


core/src/main/java/io/neow3j/protocol/core/response/ContractManifest.java
```diff
@@ -92,6 +93,24 @@ public List<ContractGroup> getGroups() {
         return groups;
     }
 
+    @JsonIgnore
+    public ContractGroup getFirstGroup() {
+        if (groups.size() == 0) {
```
gsmachado, line 98: So, I assume here that `groups` is never null because of `@JsonSetter(nulls = Nulls.AS_EMPTY)`, right?

gsmachado (approved):
Good! I believe DevEx will be enhanced with this PR. 🚀

core/src/main/java/io/neow3j/protocol/core/response/ContractManifest.java
gsmachado: This question also applies to all other attributes which has similar behavior.

mialbu: Hmm, actually, it seems the JsonSetter only initializes it as an empty array its value is `null`, but not if the value was not present during deserialization.

gsmachado: In this case, I suggest adding some tests to verify that.

gsmachado (approved):
Let's go!

PR #933, opened 2023-05-05
Feature/compiler checks

### LINKED ISSUE #861: Check for SDK methods when compiling a contract
When developing a contract and neow3j's SDK dependency is used in the same project, developers might unknowingly use methods from neow3j's SDK in a contract. This can lead to confusing exception messages that don't help the dev to locate the source of the problem.

The compiler should include checks that verify that code in a contract is not part of a package in neow3j's SDK. Then, concise compiler exceptions can be thrown.

mialbu:
Resolves #861

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
+++ b/compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
@@ -465,11 +465,35 @@ public class NeoMethod {
      */
     public void convert(CompilationUnit compUnit) throws IOException {
         AbstractInsnNode insn = asmMethod.instructions.get(0);
+        throwIfMethodHasSdkRelatedReturnOrArgType(asmMethod);
         while (insn != null) {
             insn = Compiler.handleInsn(insn, this, compUnit);
             insn = insn.getNext();
         }
         insertTryCatchBlocks();
+    }
+
+    private void throwIfMethodHasSdkRelatedReturnOrArgType(MethodNode asmMethod) {
+        Type returnType = Type.getReturnType(asmMethod.desc);
+        if (isNonDevpackNeow3jType(returnType.getClassName())) {
+            throw new CompilerException(
+                    format("The neow3j compiler does not support SDK-related types. Type '%s' used as return type of " +
+                            "method '%s' is not supported.", returnType.getInternalName(), asmMethod.name));
+        }
+        for (Type argType : Type.getArgumentTypes(asmMethod.desc)) {
+            if (isNonDevpackNeow3jType(argType.getClassName())) {
+                throw new CompilerException(
+                        format("The neow3j compiler does not support SDK-related types. Type '%s' used for an " +
+                                        "argument of method '%s' is not supported.", argType.getInternalName(),
+                                asmMethod.name));
+            }
+        }
+    }
+
+    // Returns false, if the descriptor is from io/neow3j and SDK-related, i.e., neither devpack nor compiler.
+    private boolean isNonDevpackNeow3jType(String className) {
+        return className.startsWith("io.neow3j.") &&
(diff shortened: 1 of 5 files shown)


mialbu:
I'll add another check for types in the initslot method. Hence, the mark to draft.

compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
```diff
@@ -465,13 +465,37 @@ public NeoVariable getParameterByJVMIndex(int index) {
...
     }
 
+    private void throwIfMethodHasSdkRelatedReturnOrArgType(MethodNode asmMethod) {
+        Type returnType = Type.getReturnType(asmMethod.desc);
+        if (isNonDevpackNeow3jType(returnType.getClassName())) {
+            throw new CompilerException(
+                    format("The neow3j compiler does not support SDK-related types. Type '%s' used as return type of " +
+                            "method '%s' is not supported.", returnType.getInternalName(), asmMethod.name));
+        }
+        for (Type argType : Type.getArgumentTypes(asmMethod.desc)) {
```
csmuller, line 483: Does this also cover method-local variables like the following?
`io.neow3j.types.Hash160 localHashVar = new io.neow3j.types.Hash160(...);`

csmuller (changes requested):

compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
gsmachado: Do we have tests for this? If not, we need it.

gsmachado: @mialbu what's the state of this? what @csmuller suggested/asked is possible to be done? or is already covered?

mialbu:
@csmuller After talking with @gsmachado we decided to reverse the logic.

Reason: With the previous version, we'd have had to exclude `io.neow3j.examples` as well and whenever we would provide code in any `io.neow3j.x` package, the compiler would run into a conflict with it and another exclusion (for `io.neow3j.x`) would be required in the compiler.

compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
```diff
@@ -465,13 +465,66 @@ public NeoVariable getParameterByJVMIndex(int index) {
...
+                            classOrFullyQualifiedName));
+        }
+    }
+
+    private static boolean isSDKRelatedNeow3jType(String classOrFullyQualifiedName) {
+        return isSDKRelatedNeow3jType(classOrFullyQualifiedName, "\\.");
+    }
+
+    // Returns true, if the provided name starts with 'io' and 'neow3j' and is SDK-related.
+    public static boolean isSDKRelatedNeow3jType(String classOrFullyQualifiedName, String regex) {
```
gsmachado, line 505: Why are you exposing the regex outside of this method? For me, it doesn't make sense. The regex is specific for this method implementation, and the regex will not change.

gsmachado: Moreover, it's not a regex. It's just a separator, right?

gsmachado (changes requested):
Basically just questions. Not sure yet about.

compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
mialbu: You're right, I changed the code to handle it in a more consistent way.

gsmachado (approved):
I'm satisfied with the level of this PR now. 👍

gsmachado:
@mialbu either remove @csmuller as a reviewer or he will need to approve before we merge it.

PR #936, opened 2023-05-26
Catch all exceptions and forward as an Observable when trying to fetch the latest block index

csmuller:


gsmachado:
Integration tests are running in our github-runner once again. 💪

gsmachado:
@csmuller should I merge this?

PR #943, opened 2023-07-11
Add new CryptoLib methods for ZKPs

### LINKED ISSUE #917: [Neo v3.6.0] Add new CryptoLib methods for ZKPs
https://github.com/neo-project/neo/pull/2691

mialbu:
Resolves #917 

Awaiting Neo 3.6.0 release, so we can build neo3-privatenet-docker and add int-tests for it.
Also, I haven't added any methods for it in the SDK yet. We could consider this as well.

Consider, that there might be additional opcodes added if https://github.com/neo-project/neo-devpack-dotnet/pull/726 is merged.

### TODOs
- [x] Add int-tests for all new methods (added in 728ac227)
- [x] Consider adding bls12381 methods in sdk
  ->  could also be done at a later time, imo not strictly mandatory, i.e., this should not block the compatible release for Neo v3.6.0
  -> **Decision:** we'll not include it in the SDK, since the `bls12_381` methods are intended to be used in contract code. Conracts using these methods should include separate logic (combination of add/multiply, de-/serialize of points...).

### DIFF
--- a/devpack/src/main/java/io/neow3j/devpack/contracts/CryptoLib.java
+++ b/devpack/src/main/java/io/neow3j/devpack/contracts/CryptoLib.java
@@ -65,17 +65,17 @@ public class CryptoLib extends ContractInterface {
      * Serializes a bls12381 point.
      *
      * @param g the point to be serialized.
-     * @return
+     * @return the serialized point.
      */
-    public native ByteString Bls12381Serialize(InteropInterface g);
+    public native ByteString bls12381Serialize(InteropInterface g);
 
     /**
      * Deserializes a bls12381 point.
      *
      * @param data the point as byte array.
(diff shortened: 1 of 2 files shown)


gsmachado:
@mialbu if new opcodes are added, some of the native contracts would change addresses?

mialbu:
@gsmachado The native contract address will remain the same based on the `NativeContract` class in the neo-project/neo repository. Would've been surprised if it were different tbh.

mialbu:
@csmuller 
Just fyi: There have no additional syscalls been added to the Neo core, since the functionality is all there with the new `bls12381...` methods in the native `CryptoLib` contract.

csmuller (approved):

PR #951, opened 2023-08-08
Add Sign.recoverV()

### LINKED ISSUE #949: Add a new method to get the `v` from a signature
Add a new method that allows the dev to pass in signature, public key, and message and returns the recovery value `v` for it. Then the developer can discard the public key and pass on only the signature + `v` and the message. This could be used to potentially save space.

So, basically, we would extract the following to a new (public) method:

https://github.com/neow3j/neow3j/blob/main/core/src/main/java/io/neow3j/crypto/Sign.java#L88-L102

mialbu:
Resolves #949

### DIFF
--- a/core/src/main/java/io/neow3j/crypto/Sign.java
+++ b/core/src/main/java/io/neow3j/crypto/Sign.java
@@ -168,6 +168,21 @@ public class Sign {
     }
 
     /**
+     * Given an arbitrary piece of text and an NEO message signature encoded in bytes, returns the public key that
+     * was used to sign it. This can then be compared to the expected public key to determine if the signature was
+     * correct.
+     *
+     * @param message       the encoded message.
+     * @param signatureData the message signature components.
+     * @return the public key used to sign the message.
+     * @throws SignatureException if the public key could not be recovered or if there was a signature format error.
+     */
+    public static ECPublicKey recoverFromSignature(byte[] message, SignatureData signatureData)
+            throws SignatureException {
+        return signedMessageToKey(message, signatureData);
+    }
+
+    /**
      * Finds the signature's value {@code v}, also called recovery ID, that can be used to recover the public key from
      * the signature.
      * <p>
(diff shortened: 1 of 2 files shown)


gsmachado (dismissed):
hmmm, LGTM. 👍 
🚀 🥇

gsmachado:
Let's wait for @csmuller to give his opinion.

csmuller:
@mialbu Apologies for directly fiddling with your code.... I added a second test with `v = 28` and added an overload of the recover method where you can use `SignatureData` as the sig param. This is because it is not always intuitive for the user to create two `BigInteger`s. In my test case I had to use the `BigInteger(int signum, byte[] magnitude)` constructor because otherwise `r` would have been interpreted as a negative number.

csmuller (approved):

mialbu:
> @mialbu Apologies for directly fiddling with your code.... I added a second test with `v = 28` and added an overload of the recover method where you can use `SignatureData` as the sig param. This is because it is not always intuitive for the user to create two `BigInteger`s. In my test case I had to use the `BigInteger(int signum, byte[] magnitude)` constructor because otherwise `r` would have been interpreted as a negative number.

Great additions!💯

gsmachado (approved):

gsmachado:
BTW: there's something Codacy is complaining:

https://app.codacy.com/gh/neow3j/neow3j/pullRequest?prid=12371440

Maybe consider changing that to increase code quality. 😄

core/src/main/java/io/neow3j/crypto/Sign.java
```diff
@@ -181,6 +169,67 @@ public static ECPublicKey recoverFromSignature(int recId, ECDSASignature sig, by
...
 
+    /**
+     * Given an arbitrary piece of text and an NEO message signature encoded in bytes, returns the public key that
+     * was used to sign it. This can then be compared to the expected public key to determine if the signature was
+     * correct.
+     *
+     * @param message       the encoded message.
+     * @param signatureData the message signature components.
+     * @return the public key used to sign the message.
+     * @throws SignatureException if the public key could not be recovered or if there was a signature format error.
```
csmuller, line 180: Sorry to be so picky, but the docs are a bit confusing to me. I would simplify to something like "Recovers the public key corresponding to the given signature and message."

The first `@param ` says "encoded message" is it a special encoding? if not then I would simplify to "the message".

gsmachado: Agree with @csmuller.

csmuller (approved):

PR #962, opened 2023-09-12
Fix use of PUSH0/1 instead of PUSHF/T

### LINKED ISSUE #961: Compiler uses `PUSH1` instead of `PUSHT` in boolean event parameter
The neow3j compiler uses the opcode `PUSH1` instead of `PUSHT` in the following event parameter:

```java
private static Event1Arg<Boolean> event;

public static void someMethod() {
    event.fire(true);
}
```

The JVM instructions of `someMethod` contains `ICONST_1`followed by a method call to `Boolean.valueOf()`. Neow3j converts the `ICONST_1` to `PUSH1` and ignores the following usually redundant `valueOf` method call.

The NeoVM version 3.6.0 introduced more restrictive type checks, such that when invoking `someMethod` the NeoVM throws an exception due to a type mismatch.

The specific case of JVM instructions with a `Boolean.valueOf()` should be exceptionally handled, unlike other `valueOf()` method calls. In the above case, `PUSHT` must be used.

- Compiler

mialbu:
Fixes #961

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
+++ b/compiler/src/main/java/io/neow3j/compiler/NeoMethod.java
@@ -656,13 +656,17 @@ public class NeoMethod {
      * @return true if the opcode in the last instruction is {@link OpCode#PUSH0}. False, otherwise.
      */
     public boolean lastInstructionIsPush0() {
-        OpCode opcode;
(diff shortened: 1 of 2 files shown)


csmuller (approved):
Michael Sheen Nod - 3o7aDeFfrdxWOYnx72

PR #968, opened 2023-09-13
Support RPC `findstorage`

### LINKED ISSUE #953: [Neo v3.6.0] Add new PRC `FindStorage`
https://github.com/neo-project/neo-modules/pull/805

mialbu:
Closes #953

### DIFF
--- a/int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jReadOnlyIntegrationTest.java
+++ b/int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jReadOnlyIntegrationTest.java
@@ -1257,7 +1257,7 @@ public class Neow3jReadOnlyIntegrationTest {
     public void testGetProof() throws IOException {
         long localRootIndex = 2L;
         String proof = null;
-        while (localRootIndex < 5) {
+        while (localRootIndex < 20) {
             Hash256 rootHash = getNeow3j()
                     .getStateRoot(localRootIndex)
                     .send()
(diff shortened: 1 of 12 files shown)


mialbu:
@csmuller, the integration tests are passing locally. They are failing here because the `neo3-privatenet-docker` update for Neo 3.6.0 is not yet released and the GitHub action fails to pull the image.

int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jReadOnlyIntegrationTest.java
```diff
@@ -621,6 +624,38 @@ public void testGetStorage() throws IOException {
...
 
+    @Test
+    public void testFindStorage() throws IOException {
+        NeoFindStorage.FoundStorage foundStorageActual = getNeow3j()
+                .findStorage(GAS_HASH, "", BigInteger.ZERO)
+                .send()
+                .getFoundStorage();
+
+        assertFalse(foundStorageActual.isTruncated());
+        assertThat(foundStorageActual.getNext(), is(BigInteger.valueOf(4)));
```
csmuller, line 635: @mialbu I assume you found out by experimenting that the found storage entries are truncated and that the next index is `4`. 
Do you know if the number of the entries returned in one call of this RPC method is configuratble in the node settings, e.g., like the iterator settings?

csmuller (approved):
💪 Great!

int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jReadOnlyIntegrationTest.java
mialbu: Actually, if a response is not truncated, the value `next` will directly provide you the number of existing storage entries  for the provided prefix. So, I guess that value is useful even if a response is truncated.

jiffy

csmuller: oh i see. many thanks for the explanation 🙏

PR #974, opened 2023-09-15
Support contract id in `getContractState` RPC

### LINKED ISSUE #967: [Neo 3.6.0] Support contract id parameter for RPC `getContractState`

mialbu:
Closes #967

## :rotating_light: Breaking change :rotating_light:

`ContractManagement.getContractById(int)`
👇🏼 
`ContractManagement.getContractById(BigInteger)`


csmuller (approved):
@mialbu All good, but what's the rationale behind using `BigInteger` instead of `int` for the contract ID?

mialbu:
We can also use `Integer`. `ContractIdentifiers` used `BigInteger` for the id. I'd like to be consistent, so either we change it there to `Integer` or we change the id to `BigInteger`. The point of not using `int` here is, that if the type is initialized by jackson deserialization, but somehow no value is provided for the id, the id will be `0` if we use `int`, but it should be `null`.

PR #975, opened 2023-09-15
Fix from serializable method in Transaction DTO

### LINKED ISSUE #973: `Transaction.toJson()` does not serialize transaction attribute properties
We have a method `io.neow3j.transaction.Transaction.toJson()` that transforms a `io.neow3j.transaction.Transaction` into a `io.neow3j.protocol.core.response.Transaction` DTO object in order to then serialize it to JSON. 
The method uses the `io.neow3j.protocol.core.response.Transaction(io.neow3j.transaction.Transaction tx)` constructor which in turn uses the `TransactionAttribute.fromType(...)` method to convert to transaction attributes to the `io.neow3j.protocol.core.response.TransactionAttribute` DTOs. 
But, the `fromType()` method does not set any properties on the specific attribute classes. Thus, I think the resulting JSON is missing the actual properties of the transaction attributes.

I think we never noticed, because we don't use the `io.neow3j.transaction.Transaction.toJson()` method anywhere.

mialbu:
Fixes #973 

### 1
Removed `io.neow3j.protocol.core.response.TransactionAttribute.fromType(TransactionAttributeType)`
It's replaced by the new `fromSerializable()` method
`io.neow3j.protocol.core.response.TransactionAttribute.fromSerializable(TransactionAttribute)`

### 2
`io.neow3j.protocol.core.responseOracleResponse(Integer, OracleResponseCode, String)`
👇🏼 
`io.neow3j.protocol.core.responseOracleResponse(BigInteger, OracleResponseCode, String)`

### DIFF
--- a/core/src/main/java/io/neow3j/protocol/core/response/NeoApplicationLog.java
+++ b/core/src/main/java/io/neow3j/protocol/core/response/NeoApplicationLog.java
@@ -133,6 +133,7 @@ public class NeoApplicationLog {
             return stack;
         }
 
+        @JsonIgnore
         public StackItem getFirstStackItem() {
             if (stack.size() == 0) {
                 throw new IndexOutOfBoundsException("The stack is empty. This means that no items were left on the " +
@@ -156,6 +157,7 @@ public class NeoApplicationLog {
             return notifications;
         }
 
+        @JsonIgnore
         public Notification getFirstNotification() {
             if (notifications.size() == 0) {
                 throw new IndexOutOfBoundsException("This execution did not send any notifications.");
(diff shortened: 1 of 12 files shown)


csmuller (approved):

suggalla:
When will this be released? We're running into this issue now

mialbu:
@suggalla We're working on it, and release a new version with this fix asap.

mialbu:
Hey @suggalla, we've just released a new neow3j version that supports `Neo 3.6.0` and with it the fix for #973. Check out what changed in the release notes [here](https://github.com/neow3j/neow3j/releases/tag/3.21.0).

PR #1003, opened 2023-12-08
Add StdLib.strLen()

mialbu:


mialbu:
The `neow3j-integration` job fails because the `neo3-privatenet-docker` image is not published yet for Neo `v3.6.2`.

devpack/src/main/java/io/neow3j/devpack/contracts/StdLib.java
```diff
@@ -261,4 +260,14 @@ public StdLib() {
      */
     public native String[] stringSplit(String str, String separator, boolean removeEmptyEntries);
 
+    /**
+     * Returns the length of the string.
+     * <p>
+     * Note, that the length considers UTF-16 elements as one, that means that it will return 1 for "a", "ã" and "🦆".
```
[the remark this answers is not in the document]
mialbu: Hehe yes, but I can't take the credit, they're copied from the original PR 😄

csmuller, line 266: We should not mention UTF-16 here, because i'm pretty sure that the C# NeoVM implementation uses UTF-8 encoding for strings on the VM.
E.g., the '🦆' is 4 bytes long, a Gregorian symbol like 'ა' is 3 bytes, 'ã' is 2 bytes, and 'a' is 1 byte long.
I've checked these lengths with `String.length()` which gives the length of the byte array and not the length in characters in the string.
These lengths match UTF-8 which has four "levels" for encoding characters -> 1, 2, 3, and 4 bytes.

@mialbu I'll adapt the documentation text. No need to do anything from your side.

csmuller (commented):
I will change our documentation on strings on neow3j.io. There we state for strings "The exception is the length() method, which works as expected."
Which is not actually true. If we use `.legnth()` we get the length in bytes but not in characters.

devpack/src/main/java/io/neow3j/devpack/contracts/StdLib.java
mialbu:
```
We should not mention UTF-16 here, because i'm pretty sure that the C# NeoVM implementation uses UTF-8 encoding for strings on the VM.
```

Yes, you're right. I think I didn't mention it in the SDK method but forgot to rephrase that one. Also, feel free to add `ა` in our int-tests.

devpack/src/main/java/io/neow3j/devpack/contracts/StdLib.java
```diff
@@ -261,9 +261,12 @@ public StdLib() {
...
 
     /**
-     * Returns the length of the string.
+     * Returns the number of characters of the string.
      * <p>
-     * Note, that the length considers UTF-16 elements as one, that means that it will return 1 for "a", "ã" and "🦆".
+     * Note, that this method considers each string "element" as one, that means that it will return 1 for "a", "ã" and
+     * "🦆". This length will be different from the length returned by {@link String#length()} if the string does
+     * contain non-ASCII characters. On the NeoVM, {@link String#length()} returns the byte length of the string. So,
+     * "🦆" has a byte length of 4.
```
mialbu, line 269: On the NeoVM means using the `SIZE` opcode, right?

csmuller: yes

mialbu: Could you add that, so it's clear that we're referring to the `SIZE` opcode?

csmuller: ok, added.

Issue #1008, opened 2023-12-13
Extend Transaction response object with fields for "signing context" 


csmuller:
The neo node returns not a finished transaction but a signing context (actually a `ContractParametersContext` in the C# node's code) for the following RPC methods in when it is not possible to collect all necessary signatures for the tx:
- `sendfrom`
- `sendmany`
- `sendtoaddress`
- `canceltransaction`

I didn't reproduce this yet, but I expect this to happen when the signer is a multi-sig account for which the node does not hold all corresponding private keys.
We currently don't cover this in the response objects for the above methods. Thus, code using neow3j in such a situation would fail deserialization.

PR #1009, opened 2023-12-13
Add support for `canceltransaction` RPC method.

csmuller:
There's a new issue related to this [here](https://github.com/neow3j/neow3j/issues/1008) if you're interested in some details. Though, you don't necessarily need to look at it for reviewing this PR.

### DIFF
--- a/core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
+++ b/core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
@@ -1056,11 +1056,11 @@ public class JsonRpc2_0Neow3j extends Neow3j {
      * Attempt to cancel a transaction that has been sent to the network if it is not yet confirmed. This is achieved by
      * creating a conflicting transaction (using the 'conflicts' transaction attribute). The conflicting transaction
      * has a minimally higher network fee than the transaction to cancel. You can also add an extra network fee to boost
-     * priority. Make sure that the involved signers are the same as in the transaction to cancel. The first signer will
-     * be used to pay for GAS costs. Signers can also be contracts that have a verify method.
+     * priority. At least one of the signers has to overlap with the signers on the transaction to be cancelled.
+     * The first signer will be used to pay for GAS costs.
      *
      * @param txHash   the hash of the transaction to cancel.
-     * @param signers  the signers used to perform the cancellation. The first signer will be used to pay for GAS costs.
+     * @param signers  the signers used to perform the cancellation.
      * @param extraFee an extra network fee to boost priority of the cancellation transaction. Pass the fee in fractions
      *                 of GAS. Pass null if you don't want to add an extra fee.
      * @return the request object.
--- a/core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
+++ b/core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
@@ -1059,9 +1059,9 @@ public class JsonRpc2_0Neow3j extends Neow3j {
      * priority. Make sure that the involved signers are the same as in the transaction to cancel. The first signer will
      * be used to pay for GAS costs. Signers can also be contracts that have a verify method.
      *
-     * @param txHash   The hash of the transaction to cancel.
-     * @param signers  The signers used to perform the cancellation. The first signer will be used to pay for GAS costs.
-     * @param extraFee An extra network fee to boost priority of the cancellation transaction. Pass the fee in fractions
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
```diff
@@ -1051,6 +1052,37 @@ public Request<?, NeoSendToAddress> sendToAddress(TransactionSendToken txSendTok
                 txSendToken.getValue());
     }
 
+    /**
+     * Attempt to cancel a transaction that has been sent to the network if it is not yet confirmed. This is achieved by
+     * creating a conflicting transaction (using the 'conflicts' transaction attribute). The conflicting transaction
+     * has a minimally higher network fee than the transaction to cancel. You can also add an extra network fee to boost
+     * priority. Make sure that the involved signers are the same as in the transaction to cancel. The first signer will
+     * be used to pay for GAS costs. Signers can also be contracts that have a verify method.
```
mialbu, line 1060: Do the signers need to be in the same order as well?

If so, please consider adding it here and potentially also in the `@param` javadoc entry, e.g.,:

```suggestion
     * priority. Make sure the involved signers are the same and are in the same order as in the transaction to cancel. 
     * The first signer will be used to pay for GAS costs. Signers can also be contracts that have a verify method.

File: int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jWriteIntegrationTest.java [hunk 2]
```diff
@@ -212,4 +216,43 @@ public void testSendToAddress_TransactionSendAsset() throws IOException {
...
+                .cancelTransaction(tx.getHash(), asList(Hash160.fromAddress(defaultAccountAddress())), null)
+                .send();
+
+        assertThat(cancel.getError().getMessage(), is("This tx is already confirmed, can't be cancelled."));
+    }
+
+    /*
+     * This test can have timing issues. The transaction is sent and then immediately cancelled. But, if the
+     * transaction is included in a block before the cancellation is processed, the test will fail.
+     */
```
mialbu, line 237: Is there any way to increase the block time for the NeoTestContainer that is started for this test (and this test only, i.e., not for any other test in this file)?

int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jWriteIntegrationTest.java
```diff
@@ -212,4 +216,43 @@ public void testSendToAddress_TransactionSendAsset() throws IOException {
...
+        TransactionSendToken param = new TransactionSendToken(IntegrationTestHelper.NEO_HASH, BigInteger.ONE,
+                defaultAccountAddress());
+        Transaction tx = getNeow3j().sendToAddress(param).send().getSendToAddress();
+
+        NeoCancelTransaction cancel = getNeow3j()
+                .cancelTransaction(tx.getHash(), asList(Hash160.fromAddress(defaultAccountAddress())), null)
+                .send();
+
+        assertFalse(cancel.hasError());
+        Transaction txOrCtx = cancel.getTransaction();
```
mialbu, line 249: Can you enlighten me what the context (i.e., `...OrCtx`) would include? Would that just be the original transaction in case it has already been processed?

mialbu (approved):
:100:

I've made some really minor changes and have some questions. Once answered, go ahead and merge. :muscle:

int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jWriteIntegrationTest.java
csmuller: This has to do with [this issue](https://github.com/neow3j/neow3j/issues/1008). I'll explain it in our next meeting. But, here it was just a left over that should not be there. I corrected the naming.

core/src/main/java/io/neow3j/protocol/core/JsonRpc2_0Neow3j.java
csmuller: Checked with Anna and the core code: The signers don't need to be in the same order as the 'original' transaction. Not even all of the signers need to be there. They just have to overlap with the set of signers on the 'original' transaction.

Adapting the javadoc.

int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jWriteIntegrationTest.java
csmuller: No, we set the block time in the config.json in our `test-tools` module.

PR #1018, opened 2024-01-11
Remove updatehistory field from NativeContractState

### LINKED ISSUE #1007: Remove `updatehistory` from `NativeContractState`
The `updatehistory` property has been removed from the response to `getnativecontracts` in Neo 3.6.2. Removing it from neow3j is a breaking change. Thus, we didn't yet remove it because we didn't want a new major release if it can be avoided.

Note: The `NativeContractState` DTO class will become redundant when `updatehistory` is removed.

csmuller:
...and rename its base class which is now the NativeContractState class.

Closes #1007


core/src/main/java/io/neow3j/protocol/core/response/NativeContractState.java
```diff
@@ -2,51 +2,34 @@
...
 
-    @Override
-    public int hashCode() {
-        return Objects.hash(getId(), getHash(), getNef(), getManifest());
+    public BigInteger getId() {
+        return id;
     }
 
-    @Override
-    public String toString() {
```
gsmachado, line 43: hmmm... `toString()` not required here? not even for tests, whatsoever?

csmuller: Mhhh, it wasn't there before, so I didn't notice. I'm quickly gonna voice my distaste for the `toString`, `equals`, `hashCode` methods and will add them anyway.

mialbu: Once those are added again it's good to go, imo.

gsmachado: Claude dissatisfaction when looking my comment:

Disappointed Let Down - l41YzNlhKqyJqQZG0

gsmachado (approved):
LGTM.

core/src/main/java/io/neow3j/protocol/core/response/NativeContractState.java
csmuller: 🤣

mialbu (approved):
💯

PR #1023, opened 2024-01-16
Upgrade bouncy castle targeting JDK 1.8 as min version

### LINKED ISSUE #1021: Change bouncycastle provider from bcprov-jdk15on to bcprov-jdk18on
From this:
https://central.sonatype.com/artifact/org.bouncycastle/bcprov-jdk15on/overview

To this:
https://central.sonatype.com/artifact/org.bouncycastle/bcprov-jdk18on/overview

Reason:
https://devhub.checkmarx.com/cve-details/CVE-2023-33201/

While this should not directly affect neow3j users, it might be a good time to re-evaluate if we should raise the compatibility of the neow3j SDK, from jdk 1.5 to jdk 1.8.

Since we compile to target compatibility with JDK 1.8, there's no reason why we should use `bcprov-jdk15on` instead of `bcprov-jdk18on`.

Targeting JDK 1.8 for neow3j SDK (mainly the crypto (core)) ensures compatibility with around 80+% of all Android users.

gsmachado:
Closes #1021

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext.asmVersion = '9.6'
-    ext.bouncycastleVersion = '1.70'
+    ext.bouncycastleVersion = '1.77'
     ext.grpcProtobufVersion = '1.49.0'
     ext.jacksonVersion = '2.13.4.2'
     ext.javaWebSocketVersion = '1.5.3'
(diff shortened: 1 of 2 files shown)


csmuller (approved):
👍

Issue #1016, opened 2024-01-10
Add support for setting `neow3jCompile` gradle task as cacheable or non-cacheable


gsmachado:
Gradle daemon relies on a powerful caching mechanism (yet, annoying, sometimes).

As a developer, it would be nice if I could set the neow3jCompile task as non-cacheable, for whatever reason. E.g., this would give more confidence to developers who frequently change code (in the smart contract) and would like to make sure that the compilation task is getting the latest and greatest version.

Lorenzobattistela:
Hey @gsmachado , I'm up to picking up this issue. Feel free to assign me. 
I think it is possible to change caching options through gradle API. I'll investigate it further and open a PR

gsmachado:
Cool!

As it was mentioned in Discord:

You might want to start reading some stuff from here:
https://docs.gradle.org/current/userguide/build_cache.html#sec:task_output_caching_details

The challenge is to do this programmatically, and not using groovy.
I mean, when I say "programmatically", I refer to implement this in the neow3j gradle plugin itself: https://github.com/neow3j/neow3j/tree/main/gradle-plugin

References:
- https://github.com/neow3j/neow3j/blob/main/gradle-plugin/src/main/java/io/neow3j/devpack/gradle/Neow3jCompileTask.java
- https://github.com/neow3j/neow3j/blob/main/gradle-plugin/src/main/java/io/neow3j/devpack/gradle/Neow3jCompileAction.java
- ...and other files within that directory (within the gradle-plugin module).

Lorenzobattistela:
@gsmachado tbh, I'm not really familiar with gradle, so I'm learning while researching this. I found this: https://docs.gradle.org/current/javadoc/org/gradle/caching/configuration/BuildCacheConfiguration.html
On the java API, but it seems I can only get the local build cache configuration, and not change it. I also understand that we can change it by changing the `.properties` file for gradle (that i assume is the gradle wrapper on neow3j) - https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_enable

I'm not sure if we can enable / disable this through the java api, my second thoughts on how to approach this is write to this properties file from our plugins. This seems a little odd, since touching a properties file is not ideal. 

What do you think? Maybe I'm going in the wrong direction

gsmachado:
Sure, no worries.

I would go in this direction:

[...]

PR #1026, opened 2024-01-26
Make the output of `compileJava` an input of `neow3jCompile` to have a consistent caching behavior

Lorenzobattistela:
#1016 

So @gsmachado I've followed your instructions, and I think this is the right way. I'm sure i dont understand 100% how it all happens, but Im getting along. Now I need to test this changes (specially the syntax around Spec), but I'm not quite sure how, running `./gradlew tests` or `./gradlew integrationTests` is failing, is there any setup needed? If so, where can I find it?

### DIFF
--- a/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
+++ b/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
@@ -20,11 +20,47 @@ import static org.hamcrest.Matchers.greaterThan;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
 public class Neow3jCompilePluginIntegrationTest {
 
     @TempDir
     public Path projectRootDir;
+
+    @Test
+    public void testTaskCompilationCache() {
+        String buildFileContent = "" +
+              "sourceCompatibility = JavaVersion.VERSION_1_8" + "\n" +
+              "targetCompatibility = JavaVersion.VERSION_1_8" + "\n" +
+              "\n" +
+              "neow3jCompiler {" + "\n" +
+              "    className=" + "\"io.neow3j.devpack.gradle.ContractTest\"" + "\n" +
+              "    cacheable=false\n" +
+              "}" + "\n";
+
+        GradleProjectTestCase testCase = new GradleProjectTestCase(this.projectRootDir)
+              .withDefaultDependencies()
+              .appendToBuildFile(buildFileContent)
+              .withContractName("ContractTest")
+              .withContractSourceFileName("ContractTest.java")
+              .runBuild();
+
+        BuildResult firstBuildResult = testCase.getGradleBuildResult();
+
+        assertEquals(SUCCESS, firstBuildResult.task(":" + TASK_NAME).getOutcome());
+
+        // run for the second time to assert cache was not used
+        GradleProjectTestCase testCacheNotUsed = new GradleProjectTestCase(this.projectRootDir)
+              .withDefaultDependencies()
+              .appendToBuildFile(buildFileContent)
+              .withContractName("ContractTest")
+              .withContractSourceFileName("ContractTest.java")
+              .runBuild();
+
+        BuildResult buildResult = testCacheNotUsed.getGradleBuildResult();
+        assertNotEquals(FROM_CACHE, buildResult.task(":" + TASK_NAME).getOutcome());
+        assertEquals(SUCCESS, buildResult.task(":" + TASK_NAME).getOutcome());
+    }
 
     @Test
     public void testTaskHappyPath() throws IOException {
--- a/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
+++ b/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
@@ -261,8 +261,7 @@ public class Neow3jCompilePluginIntegrationTest {
 
         // check whether the compilation miserably failed :-)
         assertEquals(FAILED, buildResult.task(":" + TASK_NAME).getOutcome());
-        assertTrue(testCase.getBuildNeow3jOutputDir().exists());
-        assertEquals(0, requireNonNull(testCase.getBuildNeow3jOutputDir().listFiles()).length);
+        assertFalse(testCase.getBuildNeow3jOutputDir().exists());
     }
 
     @Test
--- a/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
+++ b/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
@@ -30,14 +30,13 @@ public class Neow3jCompilePluginIntegrationTest {
     public Path projectRootDir;
 
     @Test
-    public void testTaskCompilationCache() throws IOException {
+    public void testTaskCompilationCacheIfFileChanges() throws IOException {
         String buildFileContent = "" +
               "sourceCompatibility = JavaVersion.VERSION_1_8" + "\n" +
               "targetCompatibility = JavaVersion.VERSION_1_8" + "\n" +
               "\n" +
               "neow3jCompiler {" + "\n" +
               "    className=" + "\"io.neow3j.devpack.gradle.ContractTest\"" + "\n" +
-              "    cacheable=false\n" +
               "}" + "\n";
 
         GradleProjectTestCase testCase = new GradleProjectTestCase(this.projectRootDir)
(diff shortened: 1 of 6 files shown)


gsmachado (changes requested):
The tests didn't run because, most likely, the code has a compilation problem. Did you try `./gradlew clean compileJava`?

Also, tests are missing. It's **rule** that for each new feature, tests should be developed. Unit tests and integration tests. For the case of `gradle-plugin` module, we just develop integration tests. [See here](https://github.com/neow3j/neow3j/tree/main/gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle).

One way is to create a test case where the `cacheable=false` also running the compilation twice. If the [task outcome](https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/TaskOutcome.html) of the second execution is `FROM_CACHE`, then the test didn't pass. If the outcome is `SUCCESS`, then it passed. We can discuss this in Discord, if you want. 👍

gsmachado:
Thinking out loud: we might think about using an `@InputDirectory` variable and setting the output of the `compileJava` task from the Java Plugin. This way we would rely on the (awesome) cache of Java Plugin already.

Lorenzobattistela:
> k abou



> Thinking out loud: we might think about using an `@InputDirectory` variable and setting the output of the `compileJava` task from the Java Plugin. This way we would rely on the (awesome) cache of Java Plugin, out of the box.

From the docs: `This will cause the task to be considered out-of-date when the directory location or contents have changed.`
So you mean basically we set the output dir as the input dir for caching, then if it has changed, the task would be considered out of date?

Lorenzobattistela:
@gsmachado At first, I've completed the initial approach so we can continue from here. I created the test case, but the second `runBuild` call is outputing `UP_TO_DATE`. I assume this happens because the files are exactly the same, so for the test I guess we have to change something for it to recompile.

gsmachado:
> @gsmachado At first, I've completed the initial approach so we can continue from here. I created the test case, but the second `runBuild` call is outputing `UP_TO_DATE`. I assume this happens because the files are exactly the same, so for the test I guess we have to change something for it to recompile.

That's a good progress!

So... let me elaborate a bit more on why I came up with [this comment](https://github.com/neow3j/neow3j/pull/1026#issuecomment-1912755788):

In the past, I was playing with neow3j and realized that, sometimes, when I changed a smart contract (i.e., a `.java` file / smart contract using the `devpack`, etc), the neow3j compiler was not being invoked -- at all. That's why I created #1016. I thought that setting a flag (e.g., `cacheable`) and configuring the `neow3jCompile` task to not be cached by the Gradle daemon would be enough. It seems it's not, as you highlighted in your test -- the `UP-TO-DATE` outcome always appears.

Thinking a bit better, I remembered about past experiments with Gradle, as well as the logic of inputs and outputs. Also, by reading the code, I realized that I went with a simpler implementation and did not properly use the output of `compileJava` being funneled into the `neow3jCompile` task.

Then, I suggest you try the following:

1. Add the following method (`replaceMethodName()`) to the `TestCaseUtils` class:

```java
    protected static void replaceMethodName(String filePath, String oldMethodName, String newMethodName) throws IOException {
        Path path = Paths.get(filePath);

        // Read the file content into a String
        String content = new String(Files.readAllBytes(path), StandardCharsets.UTF_8);

        // Replace the method name
        content = content.replaceAll("\\b" + oldMethodName + "\\b", newMethodName);

        // Write the modified content back to the file
        Files.write(path, content.getBytes(StandardCharsets.UTF_8));
    }
```

This method intends to replace a string on a file. The intention here is to change the method name of the testing smart contract. We could try to change anything else (like, the class name), but since [the smart contract used for testing](https://github.com/neow3j/neow3j/blob/main/gradle-plugin/src/test-integration/resources/ContractTest.java) is pretty simple, I'd just go with changing the method name.

2. Add a `replaceMethodName()` method to the `GradleProjectTestCase` class:

```java
    protected GradleProjectTestCase replaceMethodName(String oldMethodName, String newMethodName)
            throws IOException {
        TestCaseUtils.replaceMethodName(this.smartContractSourceFile.getAbsolutePath(), oldMethodName, newMethodName);
        return this;
    }
```

3. [In this line](https://github.com/neow3j/neow3j/pull/1026/files#diff-9bea0bc7d849b2a8a3f9c2ba3d1183e22d935f1d4c7a3cf1d6e71a0287122937R55) of your test, use the `replaceMethodName()` as, for example, `.replaceMethodName("test", "test-modified")`. As you might imagine, this intends to change the source and then run the compilation again.

4. Then, if you run the test, you will realize the following:

- [The first `buildResult`](https://github.com/neow3j/neow3j/pull/1026/files#diff-9bea0bc7d849b2a8a3f9c2ba3d1183e22d935f1d4c7a3cf1d6e71a0287122937R52) will return `SUCCESS`.
- [The second `buildResult`](https://github.com/neow3j/neow3j/pull/1026/files#diff-9bea0bc7d849b2a8a3f9c2ba3d1183e22d935f1d4c7a3cf1d6e71a0287122937R57) will return `UP-TO-DATE`.
- BUT, if you check the second `buildResult` object, more specifically checking the outcome of ALL tasks, the following will return:
image

What does this mean?

It means that, in the second `buildResult`, the `compileJava` task was not cached. This is the intended behavior since we changed the Java smart contract source (the input of the `compileJava` task), and then Gradle automatically realized that it has to recompile the smart contract to `.class` again. However, the `neow3jCompile` task simply ignored the fact that it has to recompile the smart contract. **This is a bug. It's not an intended outcome.**

Then, my initial goal to add a `cacheable` flag is maybe something we can ditch by now. We actually don't need it. I tried to fix something (that I didn't know was a bug) by introducing a feature. Not good. 😄  

So, how do we solve it? How do we make this test pass? (i.e., the second `buildResult` should return `SUCCESS` to `neow3jCompile` task after we change the source of the smart contract)

Some suggestions for you:

- Replace the whole `cacheable` thing in the `Neow3jCompileTask` class to:

```java
    @InputDirectory
    public abstract DirectoryProperty getBuildDirectory();
```

- Add the following to the end of the `Neow3jCompileTask.class` configuration:

```java
            task.getProject().getPluginManager().apply(JavaLibraryPlugin.class);
            JavaCompile compileJava = (JavaCompile) project.getTasks().getByName(JavaPlugin.COMPILE_JAVA_TASK_NAME);
            task.getBuildDirectory().set(compileJava.getDestinationDir());
            task.dependsOn(JavaPlugin.COMPILE_JAVA_TASK_NAME);
```

- Besides making sure that the `neow3jCompile` should always be triggered if smart contracts are modified, we also need to make sure that if the smart contract source does NOT change then the `neow3jCompile` should NOT be triggered. I expect you to come up with this test as well. 😸 

Happy coding.

Lorenzobattistela:
@gsmachado this other approach seems to be working well, if I'm not mistaken. The only problem I'm seeing is that one test case is breaking, probably because of directory / path stuff. Look:
`Neow3jCompilePluginIntegrationTest > testTaskCompilationError_className_notFound() FAILED`
`org.opentest4j.AssertionFailedError: expected: <true> but was: <false>`

here:
   `assertTrue(testCase.getBuildNeow3jOutputDir().exists());`

It seems that the build output directory does not exist after a compilation error.
I think this is a consequence of this lines:
```java 
JavaCompile compileJava = (JavaCompile) project.getTasks().getByName(JavaPlugin.COMPILE_JAVA_TASK_NAME);
task.getBuildDirectory().set(compileJava.getDestinationDir());
```   

Maybe they're hitting different destinations?

gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
```diff
@@ -13,19 +13,76 @@
...
+              .runBuild();
+
+        BuildResult buildResult = testCase.getGradleBuildResult();
+
+        assertEquals(SUCCESS, buildResult.task(":" + TASK_NAME).getOutcome());
+        testCase = testCase.replaceMethodName("test", "testCached");
+        // run for the second time to assert cache was not used
+        testCase.runBuild();
+        buildResult = testCase.getGradleBuildResult();
+        assertNotEquals(FROM_CACHE, buildResult.task(":" + TASK_NAME).getOutcome());
```
gsmachado, line 56: Just for clarity, I would add:

```suggestion
        assertNotEquals(FROM_CACHE, buildResult.task(":" + TASK_NAME).getOutcome());
        assertNotEquals(UP_TO_DATE, buildResult.task(":" + TASK_NAME).getOutcome());
```

not strictly required, because the following line has an `assertEquals()`, but just in case, for "maintainability" (if someone reads this test).

gradle-plugin/src/test-integration/java/io/neow3j/devpack/gradle/Neow3jCompilePluginIntegrationTest.java
```diff
@@ -13,19 +13,76 @@
...
+              .withContractName("ContractTest")
+              .withContractSourceFileName("ContractTest.java")
+              .runBuild();
+
+        BuildResult buildResult = testCase.getGradleBuildResult();
+
+        assertEquals(SUCCESS, buildResult.task(":" + TASK_NAME).getOutcome());
+        testCase.runBuild();
+        buildResult = testCase.getGradleBuildResult();
+        assertNotEquals(FROM_CACHE, buildResult.task(":" + TASK_NAME).getOutcome());
```
gsmachado, line 82: Same here:

```suggestion
        assertNotEquals(FROM_CACHE, buildResult.task(":" + TASK_NAME).getOutcome());
        assertNotEquals(SUCCESS, buildResult.task(":" + TASK_NAME).getOutcome());
```

gsmachado (changes requested):
Good work.

I requested few (minor) changes in the test files.

In addition to that, I believe that the logic of [these lines](https://github.com/neow3j/neow3j/pull/1026/files#diff-9bea0bc7d849b2a8a3f9c2ba3d1183e22d935f1d4c7a3cf1d6e71a0287122937R262-R263), which the test is failing now, can be changed -- because is not entirely correct.

I would propose to delete [these lines](https://github.com/neow3j/neow3j/pull/1026/files#diff-9bea0bc7d849b2a8a3f9c2ba3d1183e22d935f1d4c7a3cf1d6e71a0287122937R262-R263) and add this one:

```java
assertFalse(testCase.getBuildNeow3jOutputDir().exists());
```

Meaning that: if the `neow3jCompile` fails ([this line](https://github.com/neow3j/neow3j/pull/1026/files#diff-9bea0bc7d849b2a8a3f9c2ba3d1183e22d935f1d4c7a3cf1d6e71a0287122937R261)), then the build dir related to neow3j smart contracts should also not exist as a result.

Lorenzobattistela:
Done @gsmachado . Thanks for all the help. Now I understand much better how neow3j works, and I'm sure this will help on future contributions.

gsmachado (approved):
Looks good. 👍 ✅

csmuller:
@Lorenzobattistela I made a small change in the `main` branch. Please merge the neow3j `main` branch into your fork here. We're trying to make the unit tests run in fork pull-requests. Thx!

gsmachado:
@Lorenzobattistela congrats on the first PR merged to neow3j. 🎉

Issue #857, opened 2022-08-09
Add default CallFlags to contracts in devpack


mialbu:
The contracts provided in neow3j's devpack should hold default `CallFlags`. Currently, when an existing method is used from a contract class in the devpack `CallFlags.All` is used for that contract call.

Check the required call flags for all native contract methods and add them with the annotation `@CallFlags` to all corresponding methods in the devpack. Additionally, add required call flags to the provided contract interface methods, such as NEP-17 methods.

In order to use custom call flags for a provided method, the developer can extend the class and override the method with the annotation `@Override` and the desired call flags with `@CallFlags`.

Lorenzobattistela:
I'm up to pick up this one!
I'll investigate further and ask questions if needed.
Feel free to assign me as well

mialbu:
@Lorenzobattistela
The value in the annotation has to be a byte value. You can use the constants of `io.neow3j.devpack.constants.CallFlags` for this.

mialbu:
@Lorenzobattistela Also, I suggest you to read the code of the native contracts [here](https://github.com/neo-project/neo/tree/master/src/Neo/SmartContract/Native) and set the call flags in neow3j accordingly.

For example, the method `getMinimumDeploymentFee` of the native contract `ContractManagement` has call flags `ReadStates`:
https://github.com/neo-project/neo/blob/4a5a7f58567c1bf270d1c5515d90486c803b4e11/src/Neo/SmartContract/Native/ContractManagement.cs#L125-L129

Lorenzobattistela:
Yes, this is exactly what I am doing haha, I was just a little confused about the callFlags import and syntax (rusty java here). But thanks, really helpful.

mialbu:
Done in #1031

Issue #1027, opened 2024-01-26
CI-CD workflow not running when PR is submitted


gsmachado:
The CICD workflow is not running when a PR is submitted. [See this](https://github.com/neow3j/neow3j/pull/1026).

gsmachado:
@csmuller I believe this is solved. Closing.

PR #1031, opened 2024-02-12
[feat] add CallFlags for native contracts on devpack

Lorenzobattistela:
Refers to #857 

Possible reviewers:
@gsmachado 
@mialbu

### DIFF
--- a/devpack/src/main/java/io/neow3j/devpack/contracts/ContractManagement.java
+++ b/devpack/src/main/java/io/neow3j/devpack/contracts/ContractManagement.java
@@ -5,6 +5,10 @@ import io.neow3j.devpack.Contract;
 import io.neow3j.devpack.Hash160;
 import io.neow3j.devpack.Iterator;
 import io.neow3j.devpack.constants.NativeContract;
+import io.neow3j.devpack.annotations.CallFlags;
+import static io.neow3j.devpack.constants.CallFlags.AllowNotify;
+import static io.neow3j.devpack.constants.CallFlags.States;
+import static io.neow3j.devpack.constants.CallFlags.ReadStates;
 
 /**
  * Represents an interface to the native ContractManagement contract that is used to manage all deployed smart
@@ -25,6 +29,7 @@ public class ContractManagement extends ContractInterface {
      * @param hash the contract's script hash.
      * @return the contract.
      */
+    @CallFlags(ReadStates)
     public native Contract getContract(Hash160 hash);
 
     /**
@@ -33,6 +38,7 @@ public class ContractManagement extends ContractInterface {
      * @param id the contract id.
      * @return the contract.
      */
+    @CallFlags(ReadStates)
     public native Contract getContractById(int id);
 
     /**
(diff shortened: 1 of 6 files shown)


gsmachado:
@csmuller it seems that the integration tests run in a PR (and, of course, the username and password are missing/wrong -- that's why it fails).

Related to this: https://github.com/neow3j/neow3j/issues/1027

Is this the correct behavior?

mialbu:
Thanks for opening the PR, @Lorenzobattistela 💯 
I will take a look at it tomorrow.

csmuller:
> Is this the correct behavior?
@gsmachado I'm sure that the behavior is correct according to GitHub, but my understanding of it apparently is not correct.
Not sure how we can prevent the integration tests from running "fork PRs". Needs investigation.

Lorenzobattistela:
> > Is this the correct behavior?
> 
> @gsmachado I'm sure that the behavior is correct according to GitHub, but my understanding of it apparently is not correct. Not sure how we can prevent the integration tests from running "fork PRs". Needs investigation. Opened an issue.

@csmuller can you tag me in this issue? I may have some links that should help.

It seems that you can require revisions before running a workflow using `pull_request_target`. I assume this is not the wanted behavior, since we do not want to run them. 

But as per the discussion below:
https://github.com/orgs/community/discussions/26409
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

It is possible to manually disable a workflow to run by checking if the repo owner is the one you want.
It seems kinda messy but I think there's no flag for this. So if you check the repo owner and is not the original, it is a fork and then you do not run the workflow. Smth like:
```yml
jobs:
  integration_test:
    if: github.repository_owner == 'Axlabs' // run job
```

csmuller:
Hey @Lorenzobattistela here's the [issue](https://github.com/neow3j/neow3j/issues/1032) (it was living on a private project board before).

mialbu (changes requested):
There are some annotations missing. Please check and address my comments.

Also, to complete the PR, update the contracts `CryptoLib` and `StdLib` as well. They have no method with required flags (as far as I've seen from a quick look). Hence, the devpack methods should use `CallFlags.None` for these.

mialbu:
> Should we use `@CallFlags(None)` to all unspecified native functions on any contract?

I'd like that we only explicitly mark methods with call flags if the class is representing a native contract. That means, that class interfaces like e.g., `FungibleToken` or `Token` shouldn't have its methods annotated with any call flags.

I know this means that under the hood they'll still be invoked with call flags `All`. It doesn't matter if they're used on native contracts as the annotations we're making for these are mostly informative to users of neow3j (mainly so that they are aware of the feature). However, extendeable classes like `FungibleToken` shouldn't explicitly state a call flag, since it is not specified/enforced in the NEP-17 standard. So, devs should extend `FungibleToken` and override the `transfer` method if they want to use specific call flags for it.

mialbu (approved):
LGTM 💯 💪🏼

PR #1047, opened 2024-04-16
Make the "core" dependency in devpack transitive

csmuller:
Without having access to the _core_ module, developers using the devpack get warnings about undiscoverable enum values from OpCode and InteropService which are used in the devpack. Making the _core_ dependency transitive fixes the issue.

Here's how it looked like:
Screenshot 2024-04-16 at 09 55 36


gsmachado (approved):
makes sense! 🙏

PR #1050, opened 2024-04-16
Make devpack more independent of core/contract

### LINKED ISSUE #1049: Fix missing enum warnings when using the devpack
Without having access to the _core_ module, developers using the devpack get warnings about undiscoverable enum values from OpCode and InteropService which are used in the devpack. Here's how it looks like:
Screenshot 2024-04-16 at 09 55 36

Making the _core_ dependency transitive would fix the issue, but it would lead to a bad user experience, because it would make all the types from the `contract`/`core` modules available besides the `devpack` types. Many of these types are very similar, e.g., `Hash160`, but are def not interchangeable. Thus, it would be best to find another solution.

Note that with the current setup, the user is not able to use the `@Instruction` annotation properly because they don't have access to the `OpCode` nor `InteropService` enums.

csmuller:
Create copies of InteropService, StackItemType, WitnessScop, and OpCode in the devpack to make it independent of these enums in the core and make it clearer for the dev that they cannot use classes from outside of the devpack.

Closes #1049 

- [x] Add javadoc to copied enums


mialbu (commented):
Have you considered just wrapping the required values that devs need of the core in devpack classes instead of completely redefine them again in the devpack?

csmuller:
@mialbu How to wrap?

devpack/src/main/java/io/neow3j/devpack/constants/StackItemType.java
```diff
@@ -0,0 +1,14 @@
+package io.neow3j.devpack.constants;
+
+public class StackItemType {
+    public static final byte ANY = 0x00;
```
mialbu, line 4: For example, like this:

```java
public static final byte ANY = io.neow3j.types.StackItemType.ANY.byteValue();
```

mialbu: @csmuller 👆🏼

mialbu: Basically still relying on the `core` module and it's values, all while abstracting the types that are necessary and need to be disclosed to the developer.

I usually just dislike copy pasting the same values to reduce future maintenance load if something were to change. However, I know that it's likely that none of these constant values will change, so I'm not sure it's necessary. Just thinking out loud.

csmuller: Good point. I will have to check if this solution will lead to the same warnings that the dev experiences now. If not, it's a good way to make sure we have one central point where we maintain this enums and don't have real duplication.

csmuller: It doesn't work because when using `StackItemType` in an annotation (example below), then the value of the static variable (e.g., of `StackItemType.INTEGER`) must be a constant expression and cannot be a method invocation as in `byteValue`.
```
 @Instruction(opcode = OpCode.CONVERT, operand = StackItemType.INTEGER)
 ```

**Comment by csmuller** [comment]
@gsmachado I know it's not optimal from a maintenance perspective. But, note that with the current implementation, the user is not even able to use the `@Instruction` annotation properly because they don't have access to `OpCode` nor `InteropService` from the `core` module if they don't add it to the dependencies of their smart contract project.
If they do add `core` it ruins the whole dev experience that we intended for the devpack, meaning, that their IDE will show them all kinds of types from the `core` which they should never use in smart contracts.

**Comment by gsmachado** [comment]
@csmuller I changed my mind.

I think it's fine and the right way to do it. 👍 

However, I'd ask you something: can you add on top of the classes, for those that will be copied to the devpack, that this class is "duplicated" from package X (in core)? Just that we minimally document it, once new developers get onboard. You don't need to create an external documentation for this. Just add highlighted javadoc text on top of the duplicated classes.

**Comment by mialbu** [comment]
> If they do add `core` it ruins the whole dev experience that we intended for the devpack, meaning, that their IDE will show them all kinds of types from the `core` which they should never use in smart contracts.

100%, that was exactly my concern with #1047.

File: devpack/src/main/java/io/neow3j/devpack/constants/InteropService.java [hunk 2]
```diff
@@ -0,0 +1,198 @@
...
+     * Checks the signatures for the current script container.
+     */
+    SYSTEM_CRYPTO_CHECKMULTISIG,
+
+    /**
+     * Call another contract dynamically.
+     */
+    SYSTEM_CONTRACT_CALL,
+
+    // This is for internal use only.
```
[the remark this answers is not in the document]
csmuller: It's not meant to be used by smart contracts that are not native. I just added it for completeness. Will improve the comments.

core/src/main/java/io/neow3j/script/InteropService.java
```diff
@@ -40,19 +42,12 @@ public enum InteropService {
...
-     * in {@code Instruction} annotations), note, that the event name that is read from the stack MUST
-     * exist in the contract manifest's ABI. Otherwise, the NeoVM will exit with an error as soon as it reaches
-     * this syscall.
-     */
     SYSTEM_RUNTIME_NOTIFY("System.Runtime.Notify", 1 << 15),
     SYSTEM_RUNTIME_GETNOTIFICATIONS("System.Runtime.GetNotifications", 1 << 12),
     SYSTEM_RUNTIME_GASLEFT("System.Runtime.GasLeft", 1 << 4),
     SYSTEM_RUNTIME_BURNGAS("System.Runtime.BurnGas", 1 << 4),
     SYSTEM_RUNTIME_CURRENTSIGNERS("System.Runtime.CurrentSigners", 1 << 4),
+    SYSTEM_RUNTIME_GETADDRESSVERSION("System.Runtime.GetAddressVersion", 1 << 3),
```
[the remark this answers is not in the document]
csmuller: Yes, we missed that one. I'll remove it and add it in a separate PR.

devpack/src/main/java/io/neow3j/devpack/constants/InteropService.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Gets the magic number of the current network.
+     */
+    SYSTEM_RUNTIME_GETNETWORK,
+
+    /**
+     * Gets the address version of the current network.
+     */
+    SYSTEM_RUNTIME_GETADDRESSVERSION,
```
mialbu, line 152: Remove this one as well for consistency and let's include it in #1051.

mialbu (changes requested):

PR #1066, opened 2024-06-14
Improve GitHub workflow dispatch inputs

mialbu:
With these changes, the workflow dispatch input window will now look like the following:

image

image

@csmuller, I didn't get the reason for the comment on the snapshot run command on line 43 `# If you want to release a snapshot of gradle-plugin, remove the -x :gradle-plugin:publishToSonatype` since there is no `-x :gradle-plugin:publishToSonatype` there. I don't understand it in that context. Can you tell me what it's about?

Afaik, the snapshot release also publishes the gradle-plugin now (please correct me if I'm wrong). Do you think we should have an option of releasing a snapshot without the gradle-plugin? If not, I'll leave it like it is.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,10 +3,12 @@ name: Release SDK, devpack, Gradle plugin
 on:
   workflow_dispatch:
     inputs:
-      snapshot:
-        description: 'Publish snapshot?'
-        required: false
-        default: 'false'
+      release-type:
+        description: 'Release Type'
+        type: choice
+        options:
+          - Snapshot
+          - Official
 
 jobs:
   build:
@@ -21,8 +23,8 @@ jobs:
           cache: 'gradle'
       - name: Unit Tests
         run: ./gradlew --info clean test
-      - name: Release
-        if: github.event.inputs.snapshot != 'true'
+      - name: Release Official Version
+        if: inputs.release-type == 'Official'
         env:
           ORG_GRADLE_PROJECT_signingKey: ${{ secrets.GPG_KEY_ARMORED }}
           ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.GPG_KEY_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -3,10 +3,12 @@ name: Release SDK, devpack, Gradle plugin
...
-      snapshot:
-        description: 'Publish snapshot?'
-        required: false
-        default: 'false'
+      release-type:
+        description: 'Release Type'
+        type: choice
+        options:
+          - Snapshot
+          - Official
```
csmuller, line 11: Oh cool, i didn't know this.

.github/workflows/release.yml
```diff
@@ -33,14 +35,11 @@ jobs:
...
         env:
           ORG_GRADLE_PROJECT_signingKey: ${{ secrets.GPG_KEY_ARMORED }}
           ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.GPG_KEY_PASSWORD }}
           ORG_GRADLE_PROJECT_sonatypeUsername: ${{ secrets.SONATYPE_USERNAME }}
           ORG_GRADLE_PROJECT_sonatypePassword: ${{ secrets.SONATYPE_PASSWORD }}
-        run: | # If you want to release a snapshot of gradle-plugin, remove the -x :gradle-plugin:publishToSonatype
-          ./gradlew --info -x test publishToSonatype -Psnapshot 
-
-
+        run: ./gradlew --info -x test publishToSonatype -Psnapshot
```
csmuller, line 45: I probably made a mistake copying that comment or the command. The `-x :gradle-plugin:publishToSonatype` flag prevents the plugin from being published as a maven artifact. I don't think that it is of any use if it is published as a Maven artifact, only if it is published on Gradle Plugin Repo. The command in the "Official Release" workflow does that 
```
./gradlew :gradle-plugin:publishPlugin -Pgradle.publish.key=$GRADLE_PUBLISH_KEY -Pgradle.publish.secret=$GRADLE_PUBLISH_SECRET
```
Thus, I think my intention was to not publish the gradle-plugin as a Maven artifact even for snapshots.

Oooh, but then I realized that we actually do use the Gradle Plugin Snapshot as a Maven artifact, as described in `DEVELOPMENT.md`. Thus, I removed the `-x :gradle-plugin:publishToSonatype` from the command here and forgot to adapt the comment. Mystery solved. Your change is correct.

csmuller (approved):

csmuller:
In summary, we don't need a workflow that doesn't publish the plugin snapshot. Thus, this PR is ready to be merged.

gsmachado (approved):
Oh, great!
GitHub Actions added the "options" recently. It was not possible to do this before!
Looks good to me and ready to be merged.

PR #1067, opened 2024-07-29
Update test-tools for neo-express 3.7.6 changes

mialbu:
Neo express has changed how the `neo-express` file is read by default. Before v3.7.x, if no specific file was provided, the `default.neo-express` was expected in the working directory. With the release of 3.7.x, the default file is now expected in the absolute path `/root/.neo-express/default.neo-express`.

In the `neow3j-test-docker`, the `neo-express` file that is used to start a N3 blockchain is located in the absolute path `/neoxp/default.neo-express`. Thus, I adapted the `NeoExpressTestContainer` to use the `--input` flag when sending `neoxp` commands that require to be executed based on this running chain.

The behaviour of the existing methods does not change. There's no breaking change.

There is a new method `execNeoxpCommandWithDefaultConfig(String... commandParts)` which let's you send neoxp commands without the need to add the `neoxp` command as first command part, nor specify the default `neo-express` file in every call. The method is quite forgiving as it does not break if `neoxp` or the `--input` flag are used in the command parts. The input will be replaced with the default location, and the rest is ignored.


csmuller (approved):
Great!

PR #1077, opened 2024-10-15
Fix some compiler warnings

csmuller:


core/src/main/java/io/neow3j/script/VerificationScript.java
```diff
@@ -125,7 +125,11 @@ public int getSigningThreshold() {
             return 1;
         } else if (isMultiSigScript()) {
             try {
```
mialbu, line 127: This looks a bit complicated. I think we can merge the two try clauses into a single one. I had to create two separate suggestions since I can't suggest over deleted lines here in GitHub reviews.

> fyi, if you accept, you can batch the suggestions together in the `Files changed` view to create just a single commit with all suggestions.

1/2

```suggestion
            try (BinaryReader reader = new BinaryReader(script)) {
```

core/src/main/java/io/neow3j/script/VerificationScript.java
```diff
@@ -125,7 +125,11 @@ public int getSigningThreshold() {
...
-                return new BinaryReader(script).readPushInteger();
+                try (BinaryReader reader = new BinaryReader(script)) {
+                    return reader.readPushInteger();
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
             } catch (DeserializationException e) {
                 throw new RuntimeException(e);
             }
```
mialbu, line 135: 2/2

```suggestion
                return reader.readPushInteger();
            } catch (IOException | DeserializationException e) {
                throw new RuntimeException(e);
            }
        }
```

mialbu (changes requested):
Great changes! I have just one suggestion for a change. Also, please check the Codacy issues and fix them. I think it's only some minor changes there, just so we can make this check pass as well.

mialbu (approved):
LGTM 🚀

PR #1083, opened 2025-02-12
Allow combined permissions if wildcard permission has specified methods

### LINKED ISSUE #1082: Allow Combination of Permissions with method restrictions on any contract
**Description of the bug**

Neow3j `3.22+` does not allow any other permission if there's a permission with `contract = "*"` regardless of whether that permission has methods defined.

```java
@Permission(contract = "0xfffdc93764dbaddd97c48f252a53ea4643faa3fd", methods = "update")
@Permission(contract = "*", methods = "transfer")
```

**Expected behavior**

The above should be allowed as long as the permission with `contract = "*"` has methods specified.

Even though it's technically allowed to combine a `contract = "*", methods = "*"` permission with another permission, the additional permission becomes obsolete and might give the developer false promise of what's permitted. Thus, neow3j should throw a compiler exception if there's additional permissions besides a permission that allows anything.

mialbu:
Fixes #1082 

This is a non-breaking change as it expands what is allowed.

### DIFF
--- a/compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
+++ b/compiler/src/main/java/io/neow3j/compiler/ManifestBuilder.java
@@ -116,12 +116,16 @@ public class ManifestBuilder {
                 .map(ManifestBuilder::getContractPermission)
                 .collect(Collectors.toList());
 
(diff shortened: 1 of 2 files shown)


OT-kraftchain (approved):
LGTM

Issue #1081, opened 2024-12-06
Fetch block time from node when connecting


OT-kraftchain:
Required once https://github.com/neo-project/neo/pull/3612 gets merged.

gsmachado:
Check this:
https://github.com/neow3j/neow3j/blob/84675de4f1e47caa1fb9d0e0f5f99bee78c00188/core/src/main/java/io/neow3j/protocol/Neow3jConfig.java#L14

gsmachado:
Instead of hardcoding a `DEFAULT_BLOCK_TIME`, I'd suggest to check feasibility for the following:

Rely on the JSON-RPC `getversion` method to get some information from the node. Then, in there, we can find the attribute `msperblock`.

https://docs.neo.org/docs/n3/reference/rpc/getversion.html

@mialbu @OT-kraftchain

OT-kraftchain:
This feels like it should happen when connecting to the node

PR #1099, opened 2025-02-28
Add `Neow3j.init()` to fetch node-specific version info

mialbu:
Closes (#1081)

`Neow3j.build()` might be used in several places within neow3j where a node need not be present. Thus, I dislike the idea of adding an integrated RPC call in the `build()` function.

Instead, we could add another static creator function for a Neow3j object. One, that allows to fetch node-specific version info directly at object creation time. `Neow3j.init()` will do the same as `Neow3j.build()` plus additionally, it executes a `getVersion` RPC to get the node-specific protocol values and sets those required by neow3j in the Neow3j object's config. The values that can be fetched from the node and are useful for the Neow3j instance are collected in the enum `Neow3jConfig.ConfigValue`. For these values applies (a >> b meaning a preferred over b):

user-set value >> node-specific value >> default value


mialbu:
Replaced by #1103.

PR #1102, opened 2025-04-10
Add offline service for Neow3j instance

### LINKED ISSUE #1101: Create an offline implementation of Neow3j that does not require an active node connection
A neow3j instance currently needs a node address

Add a flag and an offline implementation

- SDK

mialbu:
Closes #1101

### DIFF
--- a/core/src/main/java/io/neow3j/protocol/OfflineService.java
+++ b/core/src/main/java/io/neow3j/protocol/OfflineService.java
@@ -1,0 +1,46 @@
+package io.neow3j.protocol;
+
+import io.neow3j.protocol.core.Request;
+import io.neow3j.protocol.core.Response;
+import io.neow3j.protocol.exceptions.OfflineServiceException;
+import io.neow3j.protocol.notifications.Notification;
+import io.reactivex.Observable;
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * An implementation of the Service API that restricts using any of its functions and throws appropriate exceptions
+ * if they are used.
+ */
+public class OfflineService implements Neow3jService {
+
+    private OfflineService() {
+    }
+
+    public static Neow3jService newInstance() {
+        return new OfflineService();
+    }
+
+    @Override
+    public <T extends Response> T send(Request request, Class<T> responseType) throws IOException {
+        throw new OfflineServiceException();
+    }
+
+    @Override
+    public <T extends Response> CompletableFuture<T> sendAsync(Request request, Class<T> responseType) {
+        throw new OfflineServiceException();
+    }
+
+    @Override
+    public <T extends Notification<?>> Observable<T> subscribe(Request request, String unsubscribeMethod,
+            Class<T> responseType) {
+        throw new OfflineServiceException();
+    }
+
+    @Override
+    public void close() throws IOException {
+        throw new OfflineServiceException();
+    }
+
+}
--- a/core/src/main/java/io/neow3j/protocol/exceptions/OfflineServiceException.java
+++ b/core/src/main/java/io/neow3j/protocol/exceptions/OfflineServiceException.java
@@ -1,0 +1,13 @@
+package io.neow3j.protocol.exceptions;
+
+public class OfflineServiceException extends RuntimeException {
+
+    private static final String EXCEPTION_MESSAGE = "Invalid neow3j service for this function. You cannot interact " +
+            "with a Neo node using an OfflineService instance. If you are using a Neow3j instance, make sure to " +
+            "provide a service (e.g., HttpService) that holds a valid endpoint to a Neo node when building Neow3j.";
+
+    public OfflineServiceException() {
+        super(EXCEPTION_MESSAGE);
+    }
+
+}
--- a/core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
+++ b/core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
@@ -1,0 +1,111 @@
+package io.neow3j.protocol.core;
+
+import io.neow3j.protocol.Neow3j;
+import io.neow3j.protocol.Neow3jExpress;
+import io.neow3j.protocol.Neow3jService;
+import io.neow3j.protocol.OfflineService;
+import io.neow3j.protocol.core.response.NeoGetBlock;
+import io.neow3j.protocol.exceptions.OfflineServiceException;
+import io.neow3j.protocol.notifications.Notification;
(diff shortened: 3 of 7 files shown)


core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+        OfflineServiceException thrown = assertThrows(OfflineServiceException.class,
+                () -> neow3j.getBestBlockHash().sendAsync());
+        assertThat(thrown.getMessage(), containsString(EXPECTED_OFFLINESERVICE_EXCEPTION_MESSAGE));
+    }
+
+    @Test
+    public void testNeow3jSubscribeWithOnError() throws IOException, InterruptedException {
+        AtomicBoolean onErrorExecuted = new AtomicBoolean(false);
+        neow3j.subscribeToNewBlocksObservable(true).subscribe(NeoGetBlock::getBlock, (e) -> onErrorExecuted.set(true));
+        Thread.sleep(100);
```
gsmachado, line 76: `Thread.sleep()` in tests is a [code smell](https://en.wikipedia.org/wiki/Code_smell). I know we have places where we use this.
However, let's try to not introduce more and more on our codebase.
Can we/you think of a strategy to remove this `Thread.sleep()` in all the tests here?

core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertTrue(onErrorExecuted.get());
+    }
+
+    @Test
+    public void testNeow3jCatchupAndSubscribeWithOnError() throws InterruptedException {
+        AtomicBoolean onErrorExecuted = new AtomicBoolean(false);
+        neow3j.catchUpToLatestAndSubscribeToNewBlocksObservable(BigInteger.ZERO, true)
+                .subscribe(NeoGetBlock::getBlock, (e) -> onErrorExecuted.set(true));
+        Thread.sleep(100);
```
gsmachado, line 85: same thing here.

core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+        assertTrue(onErrorExecuted.get());
+    }
+
+    @Test
+    public void testNeow3jReplayBlocksObservableSubscribe() throws InterruptedException {
+        AtomicBoolean onErrorExecuted = new AtomicBoolean(false);
+        neow3j.replayBlocksObservable(BigInteger.ZERO, BigInteger.ZERO, true)
+                .subscribe(NeoGetBlock::getBlock, (e) -> onErrorExecuted.set(true));
+        Thread.sleep(100);
```
gsmachado, line 94: and here.

core/src/main/java/io/neow3j/protocol/OfflineService.java
```diff
@@ -0,0 +1,46 @@
...
+import io.reactivex.Observable;
+
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * An implementation of the Service API that restricts using any of its functions and throws appropriate exceptions
+ * if they are used.
+ */
+public class OfflineService implements Neow3jService {
```
gsmachado, line 16: What about javadocs?! 👀

core/src/main/java/io/neow3j/protocol/exceptions/OfflineServiceException.java
```diff
@@ -0,0 +1,13 @@
+package io.neow3j.protocol.exceptions;
+
+public class OfflineServiceException extends RuntimeException {
```
gsmachado, line 3: What about javadocs? 👀

gsmachado (changes requested):
Some things to improve. 👀

core/src/main/java/io/neow3j/protocol/OfflineService.java
mialbu: I can add some, sure. We don't have any on these functions covered with javadoc btw, neither in its counterpart class `Service.java` (which is extended by HttpService and HttpService does not override Service's send function for example), nor the more frequently used functions in `Request.java` and the code there is ~7 years old. Won't mention the blame 😛😉

https://github.com/neow3j/neow3j/blob/fe11c08fca3031d2c1aa11163d7bef210b44dda1/core/src/main/java/io/neow3j/protocol/core/Request.java#L68-L78

https://github.com/neow3j/neow3j/blame/fe11c08fca3031d2c1aa11163d7bef210b44dda1/core/src/main/java/io/neow3j/protocol/Service.java#L48-L71

I will add some proper javadoc in both places. 😄

gsmachado: I know, I know.

But we should have the mentality to improve code standards bit by bit, in every PR. That's why I asked.

core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
mialbu: I'm aware it's a code smell. I haven't found a suitable way so far to wait on this disposable thread to dispose, thus, I just used sleep. I can dig a bit more.

OT-kraftchain (dismissed):

core/src/main/java/io/neow3j/protocol/exceptions/OfflineServiceException.java
mialbu: Done in e0e066c.

core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
mialbu: As communicated internally, these tests with observables are not needed in the scope of this test as they call `Neow3jService.send()` underlying which is already covered in this test class.

core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
mialbu: See https://github.com/neow3j/neow3j/pull/1102#discussion_r2037624918.

core/src/test/java/io/neow3j/protocol/core/OfflineServiceTest.java
mialbu: See https://github.com/neow3j/neow3j/pull/1102#discussion_r2037624918.

gsmachado (approved):
Great! 👍

PR #1103, opened 2025-04-11
Rework `Neow3jConfig`

### LINKED ISSUE #1081: Fetch block time from node when connecting

mialbu:
Resolves #1081 

Tests are not updated yet as parts might still change in this pull request.

I have decided to remove all node-specific protocol information from the `Neow3jConfig` class. Imo, they do not belong there. The only variable I have left statically available there is the address version since this value is used in many places via the static functions of `AddressUtils` and should not change frequently any way.

Besides the network magic and the `maxValidUntilBlockIncrement` variable, the node-specific variables do not need to be fetched, and neither be kept in the instance nor be set statically. Further, the 2 variables mentioned can be fetched from the node whenever they are actually used instead of adding this code smell and mix the client's configuration with external configuration.

In order to keep things simple and make it easy and straightforward to get the network magic (as it is used in many places) and the `maxValidUntilBlockIncrement` with a `Neow3j` instance, I have left the getters for these which now do a `getversion` RPC and return its appropriate values.

### DIFF
--- a/int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jExpressIntegrationTest.java
+++ b/int-tests/src/test-integration/java/io/neow3j/protocol/Neow3jExpressIntegrationTest.java
@@ -44,7 +44,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 @Testcontainers
-public class JsonRpc20Neow3JExpressIntegrationTest {
+public class Neow3jExpressIntegrationTest {
 
     private static final String RESOURCE_DIR = "core/";
     private static final String NEF_FILE_SOURCE = RESOURCE_DIR + "OracleTestContract.nef";


gsmachado:
Is this still "draft"?

core/src/main/java/io/neow3j/protocol/Neow3j.java
```diff
@@ -28,8 +29,8 @@ protected Neow3j(Neow3jConfig config) {
      * @param neow3jService a neow3j service instance, i.e., HTTP or IPC.
      * @return the new Neow3j instance.
      */
-    public static Neow3j build(Neow3jService neow3jService) {
-        return new JsonRpc2_0Neow3j(neow3jService, new Neow3jConfig());
+    public static Neow3j build(Neow3jService neow3jService) throws IOException {
```
gsmachado, line 32: Why does the `build()` return now an `IOException` class?

Well, I know that somewhere else you're throwing it. Just want to know why this is a good decision, architecturally speaking.

core/src/main/java/io/neow3j/protocol/Neow3j.java
```diff
@@ -131,59 +127,40 @@ public Hash160 getNNSResolver() {
...
-     * Gets the executor service used for polling new blocks from the Neo node.
-     * <p>
-     * The default executor service is a {@link ScheduledThreadPoolExecutor} with as many threads as CPUs available
-     * to the JVM.
+     * Sets the NeoNameService script hash that should be used to resolve NNS domain names.
      *
-     * @return the configured executor service.
+     * @param nnsResolver the NNS resolver script hash.
      */
-    public ScheduledExecutorService getScheduledExecutorService() {
```
gsmachado, line 141: ~as far as I know, this is required for tests. Can you please check?~

Ah, ok, I've seen you didn't delete this. Just shuffled the code around. We still have the `getScheduledExecutorService()`.

core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
```diff
@@ -513,8 +524,27 @@ public NeoInvokeScript callInvokeScript() throws IOException {
      *                                           sender cannot cover the transaction fees.
      */
     public Transaction sign() throws Throwable {
+        return sign(neow3j.getNetworkMagic());
```
gsmachado, line 527: are you using the `this.neow3j` pattern in other classes? if not, it's fine to leave like this. Otherwise, change it here.

gsmachado (changes requested):
Just some questions in this first review.

I'm really curious how neow3j will be used after these changes. 👀 
So, looking forward to the code where tests need to be adapted.

core/src/main/java/io/neow3j/protocol/Neow3j.java
mialbu: > Why does the `build()` return now an `IOException` class?

It still _returns_ a `Neow3j ` instance. It only _throws 
_ an `IOException`.

It does so because there's now an RPC happening during the creation of the returned Neow3j instance.

core/src/main/java/io/neow3j/protocol/Neow3j.java
mialbu: Correct, just reordered.

mialbu:
> Is this still "draft" because tests might change, right?

Correct, tests still need to be adapted to the changes. Once we've agreed with a design I'll update them and mark this PR as ready.

core/src/main/java/io/neow3j/protocol/Neow3j.java
gsmachado: I'm afraid that this would have a huge impact on tests and a huge refactoring of the code. Let's see.

What about throwing `Exception` instead?

core/src/main/java/io/neow3j/transaction/TransactionBuilder.java
mialbu: No, usually not.

core/src/main/java/io/neow3j/protocol/Neow3j.java
mialbu: I've created a new custom runtime exception `Neow3jBuildException` for this now.

gsmachado (approved):
Looks good! All pending questions were asked and solved.

PR #1104, opened 2025-04-11
Apply more loose default call flags in contract interfaces

### LINKED ISSUE #1098: Change default Call Flags of provided Contract Interface Functions
## Problem
The upcoming Neo version 3.8 adds additional required call flags (`AllowNotify`) to the `vote`, `registerCandidate` and `unregisterCandidate` functions of the native `NeoToken` contract. This makes the use of these functions with neow3j versions equal and newer than `3.22.+` incompatible when using these functions in the `devpack` as their default call flags will be insufficient in Neo 3.8.

## Solution Approach
Neow3j should not be too restrictive on the call flags it provides by default. If more restrictive call flags are required, these contract interface classes can always be extended and the functions can be overridden with more restrictive call flags.

Therefore, we've decided internally, that we will change the call flags of functions in contract interfaces that neow3j provides. Functions that need `WriteStates` call flags at some point will be changed to `All`, and functions that do not need `WriteStates` call flags will be changed to `ReadOnly`.

## Context
https://github.com/neo-project/neo/pull/3599

mialbu:
Closes #1098

### DIFF
--- a/compiler/src/test/java/io/neow3j/compiler/ContractInterfaceTest.java
+++ b/compiler/src/test/java/io/neow3j/compiler/ContractInterfaceTest.java
@@ -83,7 +83,7 @@ public class ContractInterfaceTest {
                 is(hexStringToByteArray(reverseHexString("ef4073a0f2b305a38ec4050e4d3d28bc40ea63f5")))); // hash
         assertThat(insns.get(22).getOpcode(), is(OpCode.NEWARRAY0)); // array, hash
         assertThat(insns.get(23).getOpcode(), is(OpCode.SWAP)); // hash, array
-        assertThat(insns.get(24).getOpcode(), is(OpCode.PUSH15)); // callflags, hash, array
+        assertThat(insns.get(24).getOpcode(), is(OpCode.PUSH5)); // callflags, hash, array
         assertThat(insns.get(25).getOpcode(), is(OpCode.SWAP)); // hash, callflags, array
         assertThat(insns.get(26).getOpcode(), is(OpCode.PUSHDATA1));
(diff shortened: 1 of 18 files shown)


compiler/src/test/java/io/neow3j/compiler/ContractInterfaceTest.java
```diff
@@ -83,7 +83,7 @@ public void testConstantStringParam() throws IOException {
                 is(hexStringToByteArray(reverseHexString("ef4073a0f2b305a38ec4050e4d3d28bc40ea63f5")))); // hash
         assertThat(insns.get(22).getOpcode(), is(OpCode.NEWARRAY0)); // array, hash
         assertThat(insns.get(23).getOpcode(), is(OpCode.SWAP)); // hash, array
-        assertThat(insns.get(24).getOpcode(), is(OpCode.PUSH15)); // callflags, hash, array
+        assertThat(insns.get(24).getOpcode(), is(OpCode.PUSH5)); // callflags, hash, array
```
gsmachado, line 86: 😱 😱 😱 😱 😱 😱 😱 😱 
what's the reason for this change?

gsmachado (changes requested):
~I just have one comment, which is somehow critical (IMHO). Without explanation, changing an opcode raises some red lights in a review. 🚨🚨🚨 😅~

All fine. 👍  🙏

compiler/src/test/java/io/neow3j/compiler/ContractInterfaceTest.java
gsmachado: oh, ok, this is the test. Sorry, this is just to accommodate the changes. All clear.

gsmachado (approved):
LGTM. 🚀

PR #1107, opened 2025-04-15
Bouncycastle and Jackson Databind Dependency Update

mialbu:
Planned for after `3.24.0`


gsmachado:
@mialbu as we talked, bouncy castle upgrade is good to go.

PR #1116, opened 2025-05-13
Policy contract updates

OT-kraftchain:


devpack/src/main/java/io/neow3j/devpack/contracts/PolicyContract.java
```diff
@@ -135,6 +137,50 @@ public PolicyContract() {
...
+     */
+    @CallFlags(ReadOnly)
+    public native int getMillisecondsPerBlock();
+
+    /**
+     * Sets the block generation time in milliseconds.
+     *
+     * @param milliseconds The block generation time in milliseconds.
+     */
+    @CallFlags(States | AllowNotify)
```
mialbu, line 151: feat: as discussed in the scope of #1104, let's use more loose call flags, i.e., `All` for setters. Devs can still extend this class, override a function and apply more strict call flags if required for some reason. Please update appropriately for all 3 new setter functions here.

devpack/src/main/java/io/neow3j/devpack/contracts/PolicyContract.java
```diff
@@ -135,6 +137,50 @@ public PolicyContract() {
...
+    public native void setMillisecondsPerBlock(int milliseconds);
+
+    /**
+     * @return the upper increment size of blockchain height (in blocks)
+     * exceeding that a transaction should fail validation.
```
mialbu, line 156: nitpick: please use the default line lenght as in the rest of the file (same for the javadoc of `setMaxValidUntilBlockIncrement`).

mialbu (changes requested):
Great work!

There's one small change I'd like to have updated in relation to call flags on native contract methods.

mialbu (approved):
🚀

PR #1150, opened 2026-01-29
Inject neow3j version to compiler at build time

### LINKED ISSUE #1149: Contract ABI reports unreleased compiler version for snapshot builds
## Problem
The ABI of a smart contract contains the compiler version that was used during compilation. However, when compiling a contract with a neow3j snapshot, this information is not correctly reflected in the ABI.

After each release, the neow3j codebase is bumped to the next version (e.g., from `3.24.0` to `3.24.1`). This version is set directly in the compiler code (e.g., in the `Compiler` class) and is therefore the version that ends up in the ABI.

The `-SNAPSHOT` suffix is only applied by the snapshot release workflow for dependency versioning, but it is not reflected in the internal compiler version. As a result, when compiling a contract using a snapshot build after `3.24.0`, the ABI records the compiler version as `3.24.1`, even though:
- 3.24.1 has not been released yet
- the contract was compiled using an unreleased snapshot build

This makes the ABI misleading, as it suggests the contract was compiled with an official release version that does not actually exist.

## Expected Behavior

If a contract is compiled using a snapshot (i.e., unreleased) compiler version, this should be reflected explicitly in the ABI.

Possible approaches include (but are not limited to):
- Appending `-SNAPSHOT` to the compiler version in the ABI
- Appending a commit hash or build identifier (e.g., `-<commit>`) to clearly indicate a non-release build

This would make it clear whether a contract was compiled using an official release or an unreleased snapshot version.

mialbu:
Fixes #1149 

I've released a snapshot that includes this change (including the neo3.9 changes). You can test it with the examples repo using the snapshot version `3.24.1-SNAPSHOT`. Compiling a contract with this snapshot will provide a NEF that has the compiler as `neow3j-3.24.1-SNAPSHOT`.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,7 +15,6 @@ at [neow3j.io](https://neow3j.io) is up-to-date.
 1. Create a branch `release` from `main`.
 2. Make sure the release version number is set in the following files:
    - `build.gradle`, in the `version` property
-   - `Compiler.java`, in the `COMPILER_NAME` constant
    - `README.md`, in the sections that explain how to import neow3j
 3. Commit and push the `release` branch, and open a pull request to `main`.
 4. Once the pull request has been merged, tag the `main` branch at the **top** commit. Tag format, e.g., `3.14.0`. 
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ ext {
 }
 
 group 'io.neow3j'
-version '3.24.1'
+version '3.24.2'
 if (project.hasProperty('snapshot')) {
     version = version + '-SNAPSHOT'
 }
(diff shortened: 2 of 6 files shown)


gsmachado (commented):
This is great and exactly what we needed.

However, have you tested this with a REAL project? I mean, a completely new project that has neow3j as a dependency and then you try to compile a smart contract with `neow3jCompile`...

If it works, give me the feedback here, and then I can approve it. 👍

gsmachado (dismissed):

mialbu:
We need to test this with an actual release (non-snapshot). I'll wait with merging this until after we've released a Neo 3.9 compatible neow3j version. I'll merge it after that to test this with a release candidate for the next version.

gsmachado (approved):
As I mentioned in the previous review, I believe this is fine.

We need to extensively test this, though, with the examples repo -- just to check if nothing is broken, and compilation works as usual.

👍

